PR #103, opened 2014-07-06
Compare history versions

### LINKED ISSUE #54: Give ability to compare two history versions
Similar to how a Wiki works, it would be very useful to be able to compare the differences between two historical versions or to compare the differences between the current version and a historical version.  The comparison tool should look at all fields that are common to both versions and show a visual diff between each pair's values.

django-reversion has an add on to give this comparison ability.  It would be great if django-simple-history had the functionality as well.

https://github.com/jedie/django-reversion-compare

macro1:
Some initial work on an admin view to compare historical versions. I was having some trouble with compatibility with 1.3 and 1.4 versions of Django admin, so for now I'm ignoring 1.3.

Will resolve #54.


jlachowski:
Hi,
what is the status of this? will it be merged into the master?

treyhunner:
I'd love to see someone recreate these changes after #277 is merged.

rossmechanic:
Closed this with an accepted issue (#396) and will put up an updated PR soon.

PR #151, opened 2014-12-27
Please support latest()

rh0dium:
This will add support for `latest()` method on history.


macro1:
I agree with @treyhunner on this... We are adding support for `latest()` without parameters, and we should be testing for that. Otherwise there could be no argument if it regressed.

I'll add some simple tests when I merge it in.

PR #152, opened 2015-02-02
add history change reason to allow reasoning for the changes

hmit:
use changeReason to set the reason. Could be None

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -26,6 +26,7 @@ Authors
 - Trey Hunner
 - Ulysses Vilela
 - vnagendra
+- hmit
 
 Background
 ==========
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ tip (unreleased)
 ----------------
 - Do NOT delete the history elements when a user is deleted.
 - Add support for ``latest``
+- Allow setting a reason for change. [using option changeReason]
 
 1.5.3 (2014-11-18)
 ------------------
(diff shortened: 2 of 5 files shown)


macro1:
I'm not completely sold on this. I need to think about it a bit more, but I would rather see some kind of 'save with additional data' function that could pass things like history user, date, and reason into the history entry. I don't really like the way we handle the other special fields either.

I really like the idea of tracking a reason with the model change though.

hmit:
The get_extra_fields(), and create_historical_record() serve kinda that function, no? I can take a dig at the save_additional_data and something like this could work:
```python
def additional_fields():
    extra_columns = [{
        'field_name': "history_date",
         'type': models.DateTimeField(),
         'attr_name': "_history_date",
         'default_value': lambda x...,
    }]
```
...etc. What do you think?

macro1:
No, the way you've added the new attribute is fine for the way we have it set up right now.

I was thinking it might be nicer to have a function that could be called to save the original model, and allow additional information to be passed along to the historical instance, rather than blindly using a signal the way we are now.

Django Activity Stream does something like this with [the way Actions are generated](https://django-activity-stream.readthedocs.org/en/latest/actions.html). There is a function that can be called directly to generate the action, or you can hook the function into a signal, capturing all saves with a less specific description.

noam-fairfly:
+1

joaojunior:
@hmit Do you continue this? If no, I will continue and make what @macro1 said.

hmit:
Hey @joaojunior, thanks for reminding me about this. 
This is a fairly old code so let me catch up with it. 
But please don't feel blocked in any way, and feel free to take it up if I don't get to it by Monday?

PR #154, opened 2015-02-09
Add history support for proxy models

macro1:
Somewhat based on https://github.com/emergence/django-simple-history/commit/9bb54005fd35ce75d887b232abb6fb44cbd13cb0.
- [ ] Test different proxy configurations (historical concrete with proxy, concrete with historical proxy, etc)

### DIFF
--- a/simple_history/__init__.py
+++ b/simple_history/__init__.py
@@ -29,5 +29,5 @@ def register(
         records.manager_name = manager_name
         records.module = app and ("%s.models" % app) or model.__module__
         records.add_extra_methods(model)
-        history_name = utils.natural_key_from_model(records.finalize(model))[1]
-        models.registered_models[natural_key] = history_name
+        records.concrete_natural_key = utils.natural_key_from_model(model._meta.concrete_model or model)
+        records.finalize(model)
--- a/simple_history/__init__.py
+++ b/simple_history/__init__.py
@@ -2,6 +2,8 @@ from __future__ import unicode_literals
 
 __version__ = '1.5.4'
 
+from . import utils
+
 
 def register(
         model, app=None, manager_name='history', records_class=None,
@@ -19,12 +21,13 @@ def register(
     `HistoricalManager` instance directly to `model`.
     """
     from . import models
-    if model._meta.db_table not in models.registered_models:
+    natural_key = utils.natural_key_from_model(model)
+    if natural_key not in models.registered_models:
         if records_class is None:
             records_class = models.HistoricalRecords
         records = records_class(**records_config)
         records.manager_name = manager_name
         records.module = app and ("%s.models" % app) or model.__module__
         records.add_extra_methods(model)
-        records.finalize(model)
-        models.registered_models[model._meta.db_table] = model
+        history_name = utils.natural_key_from_model(records.finalize(model))[1]
+        models.registered_models[natural_key] = history_name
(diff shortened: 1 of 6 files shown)


Alex131089:
Is this still on the "roadmap" ?

macro1:
Hi @Alex131089. I pulled in #112 first, as there was quite a bit more interest in allowing tracking to be inherited. From a data perspective, this support is a little odd though. Proxy models reuse the table the parent model used. If I find some time to think back through what should be happening here you'll see the progress on this branch.

Feel free to take a look at it as well of course. Pull requests are very welcome.

Alex131089:
I don't know if it can help, but for my usage, I'm using this very ugly "hack"/workaround which seems to be working : 
admin.py:
```
class HistoTrackedModel(TrackedModel):
	# /!\ Very ugly hack
	try:
		import sys
		if sys.argv[1] not in ('makemigrations',):
			raise Exception()
	except:
		history = HistoricalRecords(table_name='project_historicalTrackedModel')
  
	class Meta:
		proxy = True
		default_permissions = ('change',) # So the object appear

class HistoTrackedModelAdmin(ModelAdmin):
	readonly_fields = tuple((f.name for f in TrackedModel._meta.fields))
```
The idea here is to set an history object pointing to the historical table of the proxyfied model, except when running makemigrations because it would try to create another table for this proxy (which will not succed). No creation/modification/deletion are allowed for this proxy model, It's a readonly view of history (revert is possible & working though).
Sorry that I cannot help more, your plugin is giving a very easy way to offer history to a model :)

luzfcb:
hello @macro1, I need this feature.

I tried to base to 1.8.1, but without success. it seems that the code is no longer compatible or maybe I did something wrong.

https://travis-ci.org/luzfcb/django-simple-history/builds/117542656

kissgyorgy:
Guys, no offense, but do you even know what a proxy model is for? I suggest reading the official Django documentation: https://docs.djangoproject.com/en/1.10/topics/db/models/#proxy-models

Also if you want this "feature" because you want to track built-in models or third party models, you can already do that: https://django-simple-history.readthedocs.io/en/latest/advanced.html#history-for-a-third-party-model

This ticket should be closed IMO.

macro1:
I've been somewhat on the fence myself about adding support for proxy models. Closing, both for simplicity of the existing registry logic, and because I can't imagine a good use for it.

ThunderTrent:
I wanted to bring up a potential use case for this. I currently have an Account Data model that has several proxy models drawn from it in which all have various querysets/permissions restricting and changing various views for the django-admin. After a while of playing around with django-simple-history, everything seemed great -- but I realized later that on all of the proxy models, the saves aren't being tracked -- and the history gives an  "'Options' object has no attribute 'simple_history_manager_attribute'" error.

Perhaps I'm using Proxy models incorrectly, but it sure would of been nice for it all to work out of the box.

ThunderTrent:
Actually, I used:
register(workAccounts,table_name='web_historicalaccount')
for all the proxy models and referenced the original table and it seems to work.

macro1:
@ThunderTrent: Would something like this solve it?
https://github.com/treyhunner/django-simple-history/compare/fix-proxy-signals

I think the original ask was the ability to register proxy models... But what you're looking for is to preserve  tracking when the proxy interface is used.

silviogutierrez:
@macro1 : that's exactly what we need for our use case. Basically, make sure history is saved even when using proxy models. If there's any way I can help to get this merged, please let me know.

PR #161, opened 2015-03-20
[WIP] Add support for django < 1.8

rodxavier:
Built upon @macro1's changes. The build's currently failing on a `OperationalError: no such table` error. I'm not sure if it's some kind of a concurrency problem with sqlite or something else.


macro1:
Nice. I'll try to look at what's failing over the weekend, and try to get an idea of the impact on field type changes. It'll require a migration for anyone upgrading to Django 1.8.

PR #167, opened 2015-04-21
Force OneToOne to be a ForeignKey on historical models

### LINKED ISSUE #166: IntegrityError: Using Historical Records on an object that inherits from another non abstract object
This is just as of the latest version. I cannot edit any objects that inherit from another object.
If I create two objects as so:

```
class Zork(models.Model):
    data = models.IntegerField(blank=True, null=True, default=0)


class Bork(Zork):
    different_data = models.IntegerField(blank=True, null=True, default=0)
    history = HistoricalRecords()
```

Steps to repro:

1. Create an object of type Bork (this creates the object as well as a simplehistory record for bork)
2. Try to edit the object (this will try to create another historical record but will fail with an integrity error)

Reason is the historical record object has a OneToOne relationship to Zork, so when we try to create another historical record for zork, it throws an Integrity Error

macro1:
`OneToOneField` is really persistent about being unique. We need to treat them as `ForeignKeyField`.

Fixes #166

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Changes
 =======
 
+tip (unreleased)
+----------------
+- Fix OneToOneField transformation for historical models (gh-166)
+
 1.6.0 (2015-04-16)
 ------------------
 - Add support for Django 1.8+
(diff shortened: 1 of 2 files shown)


KevinGrahamFoster:
Awesome, thanks! :)

PR #168, opened 2015-04-22
Fix one-to-one relations in admin and restoring to historical instances

### LINKED ISSUE #162: One-to-one relations can cause models.DoesNotExist Error
If an object is created with a one-to-one relation, opportunity for breaking the History admin exists. 
To replicate:

1. Create an object to be related (RelatedA)
2. Create a SimpleHistory enabled object relating to it (ObjectB->RelatedA)
3. Create another object to be related (RelatedC)
4. Update the relation (ObjectB->RelatedC)
5. Delete the original related object (RelatedA.delete())
6. Go to "History" for ObjectB

Raises models.DoesNotExist

macro1:
I was able to reproduce and fix what @ryangallen is talking about in #162. Once I can reproduce @bradford281's original error, I'll include that before this is merged.

Fixes #162

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changes
 tip (unreleased)
 ----------------
 - Fix OneToOneField transformation for historical models (gh-166)
+- Disable cascading deletes from related models to historical models
+- Fix restoring historical instances with missing one-to-one relations (gh-162)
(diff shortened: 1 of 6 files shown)


macro1:
Looks like both were the same bug.

PR #179, opened 2015-06-29
Switch from coveralls to codecov

treyhunner:
They support branch coverage

Current coverage is [91%](https://codecov.io/github/treyhunner/django-simple-history?ref=65cfac45553842d1043473a305f92d68a5d5e962), not 93% as it says on the Coveralls badge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ install:
   - pip install -U coverage codecov $DJANGO
   - python -c 'from __future__ import print_function; import django; print("Django " + django.get_version())'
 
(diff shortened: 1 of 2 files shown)


macro1:
In practice coverage is more like 98% for what it's worth. Most of the non-coverage is from the deprecated `CustomForeignKeyField` class. I kept it in the source code for old migrations.

treyhunner:
Full disclosure: I am now working with/on Codecov

treyhunner:
That being said, merging now :smile_cat:

PR #184, opened 2015-07-07
[WIP] Generic Views Mixins

luzfcb:
It is not ready because it has no tests (sorry, I still do not know how to write tests in this time).

The idea is out of the django admin:

1 . Providing a simple way to get the historical changes of model instance in the template context with pagination (for use with DetailView, UpdateView or any view that implements the get_object() method)
2 . Providing a simple way revert model instance to a chosen version in history, for use with UpdateView (all fields of the form will be read-only) causing the revert is made in full and without modification

TODO: 
- [ ] I still would like to include an additional optional check to the revert view. Include a password field for the currently logged in User, enter your own password again to validate the revert and to avoid mistakes.
- [x] I would like to eliminate the field `all_fields`from `ReadOnlyFieldsMixin`. - fixed in https://github.com/luzfcb/django-simple-history/commit/b76dd3c1736b5233f95453ea7cd9e3252aea2750 . 
- [ ] RestoreDeletedView
- [ ] Permissions !?
- [ ] Write tests
- [ ] Write docs

any feedback is welcome.

sorry my ugly English. English is not my native language. I still have a lot to learn.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.pyc
 dist/
+.eggs
 *.egg
 *.egg-info/
 build/
@@ -12,3 +13,5 @@ test_files/
 /.ve
 /.project
 /.pydevproject
+# Pycharm project
+.idea
--- a/.landscape.yml
+++ b/.landscape.yml
@@ -1,3 +1,3 @@
-max-line-length: 79
+max-line-length: 119
 uses:
   - django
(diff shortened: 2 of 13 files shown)


rossmechanic:
Not entirely sure what this is doing. Going to create a new PR template this week and can resubmit if this is still something you want to look into.

PR #186, opened 2015-07-14
Respect the to_field setting and the target field type

arski:
https://github.com/treyhunner/django-simple-history/issues/182

Hope this is enough tests? :)

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -146,6 +146,8 @@ class HistoricalRecords(object):
                     field_arguments['db_constraint'] = False
                 if getattr(old_field, 'to_fields', []):
                     field_arguments['to_field'] = old_field.to_fields[0]
+                elif django.get_version() == "1.4" and getattr(old_field, 'to_field', None):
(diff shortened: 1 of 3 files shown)


arski:
I guess not :D

arski:
Is the code coverage test performed on a certain version of django? if that's the case then I can't really fix it.. the pre-1.6 code version would work in later versions of django too (getting the to_field from rel.field_name), but it's not really the right approach, I think at least?

Let me know your thoughts :) thanks

macro1:
Coverage should be cumulative. I'm not sure why the commit requirement is 95%, that's pretty high. I wonder if the assumption was that it's requiring 95% coverage of the new lines being added. Anyway, those are new so I wouldn't worry too much. I'll try to look at this later this week.

arski:
Yeah I'm pretty sure it's counting the new lines - since the coverage test runs on django 1.4, 1 line out of 6 new ones that is not applicable to django 1.4 is not covered by the coverage test.. (5/6 = 83.33) :)

and thanks!

PR #187, opened 2015-07-23
Add ability to edit history

buddylindsey:
We have found a need where we need the ability to have a history for a model, but we also need the ability to edit it as well. This adds in the ability to edit the history if necessary, but it can be turned on and off with a setting. By default it is off.

I still need to add docs, but just wanted to get this up.

Also I added unit tests for the feature as well. Was having trouble doing the integration tests to test this feature so opted for the unit tests.

### DIFF
--- a/simple_history/admin.py
+++ b/simple_history/admin.py
@@ -2,7 +2,6 @@ from __future__ import unicode_literals
 
 from django.core.exceptions import PermissionDenied
 from django.conf.urls import patterns, url
-from django.contrib import admin
 from django.contrib.admin import helpers, ModelAdmin
 from django.contrib.contenttypes.models import ContentType
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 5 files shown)


buddylindsey:
Finally got all the tests passing, that was a bear.

macro1:
Yeah... I have a feeling we'll be removing support for a couple Django
versions with the next minor version.

simple_history/admin.py
```diff
@@ -24,8 +24,10 @@
 
 USER_NATURAL_KEY = tuple(key.lower() for key in USER_NATURAL_KEY.split('.', 1))
 
+settings.SIMPLE_HISTORY_EDIT = getattr(settings, 'SIMPLE_HISTORY_EDIT', False)
```
macro1: I don't think we want to be modifying settings here. Could we move it into a module constant?

treyhunner: We could just remove `settings.` from all references, right?

macro1:
@buddylindsey,

I'm excited about this... It's been a while since we had a feature add. Once I get some time to try it out we'll get it merged.

buddylindsey:
Definitely check it out. 

I have added more on a local branch on my computer, to make this better. Once I get some tests written for it I'll add it to here. Will be UI fixes, more explicit that you are editing, and ability to delete history.

PR #195, opened 2015-10-10
Add history listing of deleted objects

### LINKED ISSUE #72: View deleted instances from admin page
The `SimpleHistoryAdmin` allows the history of a instance that still exists to bee seen. It would be great if there would be a way to view a list of deleted instances, and see their histories.

macro1:
It has been a really long time since I looked at this, and it turned out easier than I thought. Of course, the main admin page for the object will still give you a 404, but if you navigate directly to the history page you'll be able to view the list of changes after an object has been deleted.

resolves #72

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+sudo: false
+
 python:
   - 2.6
   - 2.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
   - DJANGO="Django>=1.8,<1.9"
 
 install:
-  - pip install -U coverage codecov $DJANGO
+  - pip install -U 'coverage<4' codecov $DJANGO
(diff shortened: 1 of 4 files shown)


KevinGrahamFoster:
Very cool. Thanks!

marcgibbons:
@macro1 Any thoughts on how best to access the list of deleted objects from the admin?

macro1:
@marcgibbons: The change above was to allow the page listing all changes to an instance (all the entries in the history table) to be visible even after the object was deleted. We don't currently offer a way to find back deleted instances easily in the admin. (So the list of changes are available, but only if you know the id of the deleted instance already.)

As always, if you find this is something useful to you, feel free to make a pull request and we can see what it would take to get it offered as a feature. If you don't feel up to the task, you can open an issue as well, describing what you're looking for.

PR #197, opened 2015-11-09
Add Django 1.9 support

macro1:
This should make everything work with Django 1.9. I also turned on warnings output to console when running the tests directly with `runtests.py`, and addressed some future warnings for 1.10. The changes didn't seem significant enough to drop 1.4 support, but we should consider doing it on the next minor release.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,7 @@ charset = utf-8
 indent_style = space
 end_of_line = lf
 insert_final_newline = true
+max_line_length = 79
 trim_trailing_whitespace = true
 
 [*.{py,rst,ini}]
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,8 @@ env:
   - DJANGO="Django>=1.9a,<1.10"
 
 install:
-  - pip install -U 'coverage<4' codecov $DJANGO
+  - pip install -U 'coverage<4' codecov
+  - pip install -U --pre $DJANGO
   - python -c 'from __future__ import print_function; import django; print("Django " + django.get_version())'
 
 script: coverage run setup.py test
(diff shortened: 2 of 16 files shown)


sgallen:
What's the status on 1.9 support and this PR? Just curious, if it's likely to be merged soon. I'm starting a new Django 1.9 project and I'm looking to include django-simple-history.

macro1:
The work is done... I just need to find some time to wrap up a release and
get it pushed into pypi

gabn88:
I still get:

simple_history/models.py:133: RemovedInDjango19Warning: Model class address.abstract_models.HistoricalUserInformation doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  type(str(name), self.bases, attrs))

When using inherit=True

If I use the registration approach (register(AbstractAddress, inherit=True)), I get

    from address.abstract_models import AbstractAddress
  File "/home/gerben/Desktop/ProductionEquinem/pre-production/address/abstract_models.py", line 175, in <module>
    register(AbstractAddress, inherit=True)
  File "/home/gerben/.virtualenvs/skillshare/lib/python3.4/site-packages/simple_history/__init__.py", line 33, in register
    records.finalize(model)
  File "/home/gerben/.virtualenvs/skillshare/lib/python3.4/site-packages/simple_history/models.py", line 88, in finalize
    history_model = self.create_history_model(sender)
  File "/home/gerben/.virtualenvs/skillshare/lib/python3.4/site-packages/simple_history/models.py", line 116, in create_history_model
    app = apps.app_configs[model._meta.app_label]
KeyError: 'address'

My Meta is:
    class Meta:
        abstract = True
        verbose_name = _('Address')
        verbose_name_plural = _('Addresses')


Or are abstract models not supported?

PR #211, opened 2016-03-13
Remove unnecessary parameter current_app

arnlaugsson:
Deprecated in Django 1.8. It's already set with:

```python
request.current_app = self.admin_site.name
```

This PR removes the warning:

> RemovedInDjango110Warning: The current_app argument of render is deprecated. Set the current_app attribute of request instead.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -27,6 +27,7 @@ Authors
 - Rajesh Pappula
 - Rod Xavier Bondoc
 - Ross Lote
+- Skuli Arnlaugsson
 - Steven Klass
 - Trey Hunner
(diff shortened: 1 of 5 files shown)


macro1:
Sorry @arnlaugsson, as you can see it's still necessary because we still support 1.6 and 1.7. If you can find a clean way to resolve the warning without removing the backward compatibility, I'm open to it. You could also look into silencing the warning, if you're just trying to cleanup your project's test output.

arnlaugsson:
Thanks - yeah, I saw the same when I ran in on older Django versions. I'll close for now at least.

PR #212, opened 2016-03-14
flake8 fixes

robmoorman:
some small flake8 fixes


macro1:
I appreciate your enthusiasm for cleaning up our code, but I'm going to go ahead and close this. The readability of the code in django-simple-history, and this holds true for many projects, suffers greatly from the organization of the code. These newlines don't really factor in.

PR #229, opened 2016-07-22
Allow filtering of as_of

pbrooks:
Since as_of is a model level method, allowing filtering within the queryset.

### DIFF
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -56,7 +56,7 @@ class HistoryManager(models.Manager):
                                              self.instance._meta.object_name)
         return self.instance.__class__(*values)
 
-    def as_of(self, date):
+    def as_of(self, date, **kwargs):
         """Get a snapshot as of a specific date.
 
         Returns an instance, or an iterable of the instances, of the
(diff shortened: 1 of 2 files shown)


macro1:
I really like the idea of this PR. What might be more helpful though is to change the history manager into a history queryset. The queryset could then be used as a manager, allowing the utility methods to be available on any queryset of historical models.

leportella:
I like the suggetion from @macro1 but I think that this PR seems ok for now. What do you guys think? @treyhunner

treyhunner:
Hm at first I thought we could add @macro1's suggestion later, but then I realized what this actually adds.

This would allow:

    Poll.history.as_of(time, question="what's up?")

But I do think we might want to instead support:

    Poll.history.filter(question="what's up?").as_of(time)

It seems more clear what that is querying to me. I'll see if I can put together a pull request for that.

treyhunner:
So the change was relatively simple, but we'll want to update the tests for `as_of` and `most_recent` to test for the queryset behavior as well: https://github.com/treyhunner/django-simple-history/compare/manager-to-queryset

It looks like the tests are in two places for this model manager (strangely): `test_models.HistoricalRecordsTest` and `test_manager.AsOfTest`. Those should probably be merged eventually.

Also it looks like this change doesn't work on Django 1.6. We should probably drop Django 1.6 and Python 3.3 support anyway...

rossmechanic:
Closing with accepted issue #397. Will put up updated PR soon.

Issue #202, opened 2015-12-02
Add a 'batch_size' option to the populate history command


macro1:
As mentioned in #189 it can be problematic to run the populate history command on very large data sets. A parameter to the populate history command should be added to allow the population to be performed in chunks.

rossmechanic:
Fixed in #263

Issue #189, opened 2015-07-31
SSL SYSCALL error


psychok7:
Hi there so i am trying  django-simple-history + django 1.7 and i have a table with about 300.000 records. When i run `populate_history` i get an error:

```
(xxx)vagrant@vagrant:/vagrant$ python manage.py populate_history --auto --settings=xxx.settings.local
INFO 2015-07-31 14:54:39,572 base 1713 139923734366016 Raven is not configured (logging is disabled). Please see the documentation for more information.
Existing history found, skipping model <class 'support_tables.models.ParishCouncil'>
Existing history found, skipping model <class 'support_tables.models.Occupation'>
Saving historical records for <class 'support_tables.models.PostalCode'>
Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/home/vagrant/.virtualenvs/xxx/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/home/vagrant/.virtualenvs/xxx/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 377, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/vagrant/.virtualenvs/xxx/local/lib/python2.7/site-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "/home/vagrant/.virtualenvs/xxx/local/lib/python2.7/site-packages/django/core/management/base.py", line 338, in execute
    output = self.handle(*args, **options)
  File "/home/vagrant/.virtualenvs/xxx/local/lib/python2.7/site-packages/simple_history/management/commands/populate_history.py", line 61, in handle
    self._process(to_process)
  File "/home/vagrant/.virtualenvs/xxx/local/lib/python2.7/site-packages/simple_history/management/commands/populate_history.py", line 106, in _process
    utils.bulk_history_create(model, history_model)
  File "/home/vagrant/.virtualenvs/xxx/local/lib/python2.7/site-packages/simple_history/management/commands/_populate_utils.py", line 27, in bulk_history_create
    history_model.objects.bulk_create(historical_instances)
  File "/home/vagrant/.virtualenvs/xxx/local/lib/python2.7/site-packages/django/db/models/manager.py", line 92, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
[...]

PR #231, opened 2016-08-01
Added batch size option to the management command for populating the history

mikeengland:
Currently, if you try and populate the history of a large table Django raises a 'lost connection to MySQL server' error. This is because the bulk_create method is not given a batch size.

This pull request adds an additional parameter to the management command (--batchsize), where you can specify this e.g. 500. I have set the default value to a reasonable 200. Issue #202 and #189 should be fixed in this pull request.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -33,6 +33,7 @@ Authors
 - Ulysses Vilela
 - vnagendra
 - Lucas Wiman
+- Michael England
 
 Background
 ==========
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changes
 tip (unreleased)
 ----------------
 - Add Polish locale.
+- Added --batchsize option to the populate_history management command.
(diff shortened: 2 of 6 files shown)


mikeengland:
Hmm... looks like some of the TravisCI builds have failed due to the error `OperationalError: no such table: tests_historicaltrackedwithmultipleabstractbases`. Has anyone seen this before?

macro1:
Yes, I've seen errors like this before. There are some race conditions that can surface because we're dynamically creating models within some of the tests.

The table giving you trouble here is correctly missing... I'm not sure why Django thinks it should exist though. Here's the code in the tests: https://github.com/treyhunner/django-simple-history/blob/1.8.1/simple_history/tests/tests/test_models.py#L801

mikeengland:
Ok, thanks @macro1 !

mikeengland:
Please can someone take a look at merging this? It would be very useful to have in master.

PR #233, opened 2016-08-04
Django 1.10 compatibility for admin.py

flesser:
Parameters for `render()` have changed (deprecated in 1.8, removed in 1.10):
- `dictionary` was renamed to `context`
- `current_app` was removed, instead `request.current_app` should be set.

I have not tested with older releases yet, but according to the docs, it should work down to 1.8 when the changes were introduced.

### DIFF
--- a/simple_history/admin.py
+++ b/simple_history/admin.py
@@ -76,7 +76,7 @@ class SimpleHistoryAdmin(admin.ModelAdmin):
         }
         context.update(extra_context or {})
         return render(request, template_name=self.object_history_template,
-                      dictionary=context, current_app=request.current_app)
+                      context=context)
 
     def response_change(self, request, obj):
         if '_change_history' in request.POST and SIMPLE_HISTORY_EDIT:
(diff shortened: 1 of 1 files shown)


macro1:
Hello @flesser, thanks for your work on this. Unfortunately we maintain backward compatibility with older versions of Django, so we cannot merge your changes the way they are now.

If you like, you can build off of some work I did  while back:
https://github.com/treyhunner/django-simple-history/compare/django110
Latest test run: https://travis-ci.org/treyhunner/django-simple-history/builds/149779056

 I had the tests passing with 1.8 through 1.10. I think I was considering dropping support for 1.6 and 1.7, because the differences in how options in Django commands are defined become a bit of a pain.

triat:
Actually it should work on the django 1.7 version (the oldest version supported by django right now)

I just had a quick read on that but related to that, if you don't set it, it's gonna take the default one, it's what we want right? 
[https://docs.djangoproject.com/en/1.7/topics/http/urls/#topics-http-reversing-url-namespaces](https://docs.djangoproject.com/en/1.7/topics/http/urls/#topics-http-reversing-url-namespaces)

macro1:
Resolved in #259, including backward compatibility with 1.6 and adjusted tests.

PR #236, opened 2016-08-16
exclude some fields from history models. 

maparent:
Solves the first part of issue 165.
Open to suggestions of needed tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changes
 tip (unreleased)
 ----------------
 - Add Polish locale.
+- Add a Meta option exclude_from_history, which can recevie a collection of field names.
 
 1.8.1 (2016-03-19)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changes
 tip (unreleased)
 ----------------
 - Add Polish locale.
-- Add a Meta option exclude_from_history, which can recevie a collection of field names.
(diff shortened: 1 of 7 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,7 @@ Changes
 tip (unreleased)
 ----------------
 - Add Polish locale.
+- Add a Meta option exclude_from_history, which can recevie a collection of field names.
```
luzfcb: typo error:
`receive`

and not `recevie`

maparent: thank you, corrected.

simple_history/models.py
```diff
@@ -307,5 +318,5 @@ def __init__(self, model):
 
     def __get__(self, instance, owner):
         values = (getattr(instance, f.attname)
-                  for f in self.model._meta.fields)
+                  for f in HistoricalRecords.get_fields(self.model))
```
macro1: Could we change this one to use `super()`? That would allow others to override `get_fields()` without needing to override `__get__()` as well.

maparent: Good point nonetheless. I found another case I had missed, in HistoryManager.most_recent.
I realize I do not know how to get the HistoricalRecord object from the instance; I'm now thinking the get_fields method should actually be in the HistoryManager and not the HistoricalRecord. Would you agree?

maparent: OK, I did move get_fields. Tell me if this suits you.

macro1:
Closing in favor of #238

PR #238, opened 2016-09-12
Exclude fields

joaojunior:
Create option to pass fields to not copy in history


gyermolenko:
@joaojunior hi. AFAIK using mutable objects in arguments is not desired.
[http://stackoverflow.com/questions/4535667/python-list-should-be-empty-on-class-instance-initialisation-but-its-not-why](url)

macro1:
Agreed, this is perfect except for the mutable default parameter. The issue being every instance of the class with have the exact same list as the default list of excluded fields. As the code is now, that default list is never used. But if it were, it would be changed for all instances using the default exclusion list. I'll change it in the merge commit I guess.

joaojunior:
I fix the errors found here and I send a new PR to this: #274

PR #246, opened 2016-11-10
Check that request.user is a user model that we can save

### LINKED ISSUE #245: History creation fails if request.user isn't an AUTH_USER_MODEL
We have views where `request.user` is not the same as AUTH_USER_MODEL which causes the history model creation to blow up.

`get_history_user` assumes that `request.user` is always saveable into the history record - it should do an explicit `isinstance` check to make sure.

mbox:
If request.user isn't the same as AUTH_USER_MODEL for some reason, then injecting the user into the history model will fail.

Fixes #245

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -8,6 +8,7 @@ from django.db import models, router
 from django.db.models.fields.proxy import OrderWrt
 from django.conf import settings
 from django.contrib import admin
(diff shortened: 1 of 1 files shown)


rossmechanic:
Going to close this since we're going to merge #371 soon, and that should fix your issue.

PR #255, opened 2016-12-20
Added edit_history option to HistoryManager.as_of(). 

wizpig64:
Currently, the history manager's `as_of` method returns instances of the original model type. This means when trying to edit history, filtering and dealing with DoesNotExist errors has to be done manually. This adds getting HistoricalObject instances as of a certain time right in.

From the docstring:

> When edit_history is enabled, historical objects (which can be updated
> to modify history) are returned, rather than instances of the original
> object type. Otherwise, saving will create new history objects.

Not sure how I would set up a test for the returned instance's class, mainly because Historical<Object> classes are defined on the fly. Feedback welcome, thanks.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -23,7 +23,12 @@ as_of
 ~~~~~
 
 This method will return an instance of the model as it would have existed at
-the provided date and time.
+the provided date and time. Modifying and saving this instance will revert the
+parent instance to the instance snapshot you are saving, creating a new
+historical record at the current time.
+
+Specifying ``edit_history=True`` returns the equivalent historical object
+record, which can be deleted or updated to edit history.
 
 .. code-block:: pycon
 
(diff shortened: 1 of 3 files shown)


treyhunner:
I'm having trouble imaging the purpose of editing historical records. What's the use case for this?

wizpig64:
@treyhunter Here's my situation:

(By the way, if you think I'm not using django-simple-history in a safe way or should otherwise be doing something different, let me know.)

Our business generates reports for farms every few weeks, and bills by the acre. We want to be able to render old reports correctly, as well as bill appropriately given the correct data at the time. HistoricalRecords lets us keep track of changes and render relational data as of the report's date, which is great! If a farm drops a few acres mid-season or switches crops between years, the older reports are still correct because they show the details *at the time of the report*.

This solution without the ability to edit history works fine as long as we have perfect information from our growers. But sometimes crops can change and not be noted in the database, and nobody will notice that anything is wrong until we audit the reports at the end of the month. So being able to go back in time and change something *as of the correct date* is pretty important to us; it lets us correct our mistakes.

Right now, saving the output of as_of(date) simply creates a new historical record with the current date, [just as it does in the admin](https://github.com/treyhunner/django-simple-history/issues/254). This pull request would make it easier to implement editing history in my editor, as well as the admin if that were to be desired behavior in the future.


Thanks.

PS: Looking back at this 6-month-old PR, it might be cleaner to implement this as a seperate method, like edit_as_of, rather than confusing things with flags. If you're open to the idea of editing historical records, I'll give it another go.

rossmechanic:
I see django-simple-history more as an audit log with the ability to revert, where we'd want it to be difficult to go back and edit history. Thus, I don't really see this feature as part of what django-simple-history should be accomplishing. However, we could have a setting that allows history to be editable. Going to close this for now. If you want to push on it, feel free to add an issue and we can discuss there.

PR #256, opened 2017-01-03
Field list

gbataille:
in the same way that django admin propose a `list_display` array of fields to display in the object list view, it would be nice to have a `history_list_display` array of fields to display in the object history list view. That way, we could track easily which field has changed (typically one important like an object status) without having to open the details one by one.

Seems to do exactly what I want when I pull it in my project. It's way less complex than the django admin object list, but I think the complexity is not needed here (or could be added later).

_Note:_ I'm in the mountain with limited internet. I was unable to run the test suite (I still coded one), maybe because it's trying to run it against django 1.10 that is not supported. In any case, I haven't looked further into testing it yet. Let me know if you find issues.


gbataille:
There are errors on 3 builds in travis, but they don't seem to be related to my changes
> ```OperationalError: no such table: tests_historicaltrackedwithmultipleabstractbases```

I'll let you advise

gbataille:
any comments ? should I simply resolve the conflicts?

gbataille:
I upped the code coverage.
Can you let me know what you think about it?

Thanks

simple_history/templatetags/getattributes.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+
+register = template.Library()
+
+
+def getattribute(value, arg):
+    """Gets an attribute of an object dynamically from a string name"""
+
+    if hasattr(value, str(arg)):
+        return getattr(value, arg)
```
rooterkyberian: why do you call `str` on `arg` in `hasattr`? you don't do it with `getattr`
also you don't need that `if hasattr(...):` in the first place
```py
def get_attribute(value, arg):
    return getattr(value, arg, None)
```

gbataille: True, dunno, must have evolved that way :)

Fixed

treyhunner:
I think I understand what this change does but I'm not certain. Seeing a screenshot of this would be helpful.

I would also be open to adding a screenshot to the documentation if you think it makes sense.

gbataille:
Here it is:
The visualization of the history is pretty dull by default. You don't get to see what has changed. And in a model that changes quite a bit, finding a point in time where some changes were made is challenging.
screen shot 2017-06-10 at 10 07 42


With the change, you can control which columns are visible in the history list view, and therefore get a quick glimpse at when some information was modified
screen shot 2017-06-10 at 10 06 05
screen shot 2017-06-10 at 10 07 00


I'll see to add that to the documentation

gbataille:
and here you are

treyhunner:
@gbataille cool! This change looks good to be. 👍

I think this will be good for merging after the current merge conflicts are resolved with master.

gbataille:
done

treyhunner:
Thanks @gbataille!

PR #258, opened 2017-01-17
Specify compatibility through Django 1.9.x only

shanemgrey:
Django 1.10 has been out over 5 months and most new projects are probably using it.  It's worth calling out clearly that the breaking changes in Django have caused django-simple-history to be incompatible so that users know ahead of time and can decide to downgrade Django, update this repo, or not use it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,9 +18,10 @@ django-simple-history
    :alt: Code Health
 
 
+
 django-simple-history stores Django model state on every create/update/delete.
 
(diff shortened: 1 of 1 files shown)


macro1:
@shanemgrey: Thanks for taking the time to make this PR. Fortunately I don't think we'll need it. I was having some issues with some race conditions in the tests, but I think it's sorted. We should have a release with 1.10 compatibility shortly! master is already compatible.

shanemgrey:
That's great to hear! Thank you for fixing the issues.  That's a far better commit.  :-)

Issue #257, opened 2017-01-05
Test is broken in master 


leportella:
I just run the tests in the master branch and I am getting this error:

![image](https://cloud.githubusercontent.com/assets/7255297/21688726/1672544e-d355-11e6-9f4b-8e02c953e234.png)

macro1:
@leportella, I'm sorry the tests weren't running for you. The error you're seeing is actually related to a Django 1.10 incompatibility that was in master. I just merged my work on adding Django 1.10 compatibility, so running the tests in master against 1.10 should work now.

Hopefully you find some time to revisit checking out this project!

Issue #248, opened 2016-11-18
HistoryRequestMiddleware not defined


jhotujec:
Following official guide I get this error:

`django.core.exceptions.ImproperlyConfigured: WSGI application 'optilimo.wsgi.application' could not be loaded; Error importing module: 'Module "simple_history.middleware" does not define a "HistoryRequestMiddleware" attribute/class'`

Any ideas how to fix this? Thanks!

macro1:
I believe this is related to a change in the middleware API in Django 1.10. That should be addressed in master now. I'm going to optimistically close this, please do reopen if the changes in master don't resolve the issue you're seeing.

GoudekettingRM:
Sorry to open up this old issue, but if I add the app like so:

```py
INSTALLED_APPS = [
    ....,
    "simple_history",
    ....
]
```
and the middleware like so:
```py
MIDDLEWARE = [
    ....,
    "simple_history.middleware.HistoryRequestMiddleware",
]
```

I get this same error once more:

```
 ImproperlyConfigured(
django.core.exceptions.ImproperlyConfigured: WSGI application 'rocketfuel.wsgi.application' could not be loaded; Error importing module.
```

```
Django: 3.0.3.
DjangoSimpleHistory: 3.5.0
Python: 3.9.16
```

ddabble:
@GoudekettingRM We don't support Django versions below 3.2 (see the supported versions in the README or in [the docs](https://django-simple-history.readthedocs.io/en/latest/)). Does the issue persist if you upgrade Django to one of the supported versions - preferably 5.0? 🙂

Issue #241, opened 2016-10-13
I try usage in Django 1.10.2 and I am seeing error in runserver


lukasgarcya:
In my project I'm using Django 1.10 and when I run runserver does not work due to the error below:

Unhandled exception in thread started by <function wrapper at 0x7f2211a89668>
Traceback (most recent call last):
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/utils/autoreload.py", line 226, in wrapper
    fn(_args, *_kwargs)
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/core/management/commands/runserver.py", line 142, in inner_run
    handler = self.get_handler(_args, *_options)
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/contrib/staticfiles/management/commands/runserver.py", line 27, in get_handler
    handler = super(Command, self).get_handler(_args, *_options)
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/core/management/commands/runserver.py", line 64, in get_handler
    return get_internal_wsgi_application()
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/core/servers/basehttp.py", line 49, in get_internal_wsgi_application
    return import_string(app_path)
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/utils/module_loading.py", line 20, in import_string
    module = import_module(module_path)
  File "/usr/lib64/python2.7/importlib/**init**.py", line 37, in import_module
    **import**(name)
  File "/home/lukasgarcya/workspace/django_atual/django_atual/wsgi.py", line 16, in <module>
    application = get_wsgi_application()
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/core/wsgi.py", line 14, in get_wsgi_application
    return WSGIHandler()
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 153, in **init**
    self.load_middleware()
  File "/home/lukasgarcya/.virtualenvs/django_atual/lib/python2.7/site-packages/django/core/handlers/base.py", line 82, in load_middleware
    mw_instance = middleware(handler)
TypeError: object() takes no parameters

nessalc:
[...]

Issue #237, opened 2016-08-27
django-simple-history plugin doesn't work correctly with django 1.10 admin UI


leonmaks:
When django-simple-history was added to replace default admin UI in django 1.10 the following exception was raised: Exception Value: render() got an unexpected keyword argument 'dictionary' 

Please use the following link for details: http://stackoverflow.com/questions/39179559/django-simple-history-plugin-render-got-an-unexpected-keyword-argument-dictihttp://stackoverflow.com/questions/39179559/django-simple-history-plugin-render-got-an-unexpected-keyword-argument-dicti

leonmaks:
I've corrected this issue by replacing 2 calls of **render** function in simple-history.admin module:
1:
```
78        return render(request, template_name=self.object_history_template,
79                      context=context)

```2:
```
181        return render(request, template_name=self.object_history_form_template,
182                      context=context)
```

macro1:
django-simple-history isn't compatible with 1.10 yet. I have a branch in progress, but work isn't complete.

akhayyat:
What other incompatibilities with Django 1.10 are there, besides the admin integration and the middleware settings?

macro1:
There's also a change in the way management commands are defined, we may be dropping 1.6 support due to that. But it's not too bad over-all, I just haven't had time to work on it yet.

Here is what I have in progress right now: https://github.com/treyhunner/django-simple-history/compare/django110

KevinGrahamFoster:
Hey @macro1 is there anything I can do to help you with this? What remains to be done? I can jum on finishing it up if you want (and if I'm able...)

achillis1:
Any update on this issue? I really need it to be compatible with 1.10...

nessalc:
I took @macro1's django110 branch and made a slight change to `middleware.py`, available [here](https://github.com/nessalc/django-simple-history/blob/django110/simple_history/middleware.py). It works for 1.10 and the current dev branch of Django, at least, but may not be backward (or forward) compatible from there.

achillis1:
I'm still new to python django. @nessalc, could you please elaborate how to install your fix? I'll need it to be working on heroku too.

nessalc:
@achillis1: If you run the following:

    git clone -b django110 https://github.com/nessalc/django-simple-history.git
    pip install django-simple-history/

[...]

PR #259, opened 2017-01-18
Add Django 1.10 support

### LINKED ISSUE #234: MIDDLEWARE_CLASSES setting is deprecated in Django 1.10
Hello there.

I am new to django and django-simple-history. I am testing this app and, after following Quick Start and making migrations, I got this warning:

`$ python manage.py makemigrations myapp`
`System check identified some issues:`
`WARNINGS:`
`?: (1_10.W001) The MIDDLEWARE_CLASSES setting is deprecated in Django 1.10 and the MIDDLEWARE setting takes precedence. Since you've set MIDDLEWARE, the value of MIDDLEWARE_CLASSES is ignored.`

How can the Quick Start guide be changed to use MIDDLEWARE instead of MIDDLEWARE_CLASSES? Or should there be another solution, besides downgrading django to at most 1.9 ?

Thank you very much for your time and effort.

macro1:
Made changes made necessary by Django 1.10:

- Support new middleware API
- Fix admin view paramaters
- Update management command definition

Also split the tests involving model registry into their own app, effectively a separate suite. This prevents partially defined models from interfering with other tests.

Resolves #234, #237, #241, #248, #253, #257


macro1:
I realized after I merged this I missed the documentation changes... Those will be coming shortly, and hopefully a release in the next couple days. The big blocker was getting tests in a stable state (there were race conditions with the model definition tests).

PR #266, opened 2017-03-02
Isn't it logical to not show history for someone without access?

ambivalentno:


treyhunner:
Thanks @ambivalentno!

If someone would like to add tests for permission success/failure case, I think this will make sense to merge.

PR #276, opened 2017-05-19
added py36 and Django 1.11 to test environments

rooterkyberian:
* fixed up tests so they work with Django 1.11
* added python 3.6 as test enviroment
* removed python <3.5 versions from tox for django-trunk as next Django version (2.0) won't support anything less https://docs.djangoproject.com/en/1.11/faq/install/#what-python-version-can-i-use-with-django

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - 3.3
   - 3.4
   - 3.5
+  - 3.6
 
 env:
   - DJANGO="Django>=1.6,<1.7"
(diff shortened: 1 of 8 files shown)


macro1:
Did you want to add yourself to [AUTHORS.rst](https://github.com/treyhunner/django-simple-history/blob/master/AUTHORS.rst) as well?

rooterkyberian:
sure, done

clintonb:
@macro1 any plans to merge/release this?

leportella:
This seems good! It would be better to add the support on documentation, but I think we can leave it to another PR. Thank you!

clintonb:
@leportella any chance you can create a new release on PyPI?

PR #277, opened 2017-05-27
Added Model Delta and HistoricalChanges

### LINKED ISSUE #244: Add a "ModelDelta" and "__sub__" method to have a timeline info
We needed to implement a timeline considering the changes made in a specific model. For that, we basically needed the "difference" between two records.

In order to do that, we created a `__sub__` method that allowed us to simply do something like this:
`history[1] - history[0]`. 

The result of this operation is an instance of a class called ModelDelta, that gives us the following information: 
   * the new value 
   * the old value 
   * the fields that changed in the process
   * user that made the change (to the new value)

On this ModelDelta class we implemented a function `to_timeline` that get every changed field and apply a function to create a timeline entry with some messages like `This user changed this fields at this hour`.

I would like to know if you would think this is interesting to be a feature in the project, and if you have suggestion for better approach.

leportella:
This is related to the issue #244 that I have opened earlier.

Hope it is helpful :)

Closes #244

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -44,7 +44,7 @@ class HistoricalRecords(object):
         try:
             if isinstance(bases, six.string_types):
                 raise TypeError
-            self.bases = tuple(bases)
+            self.bases = (HistoricalChanges,) + tuple(bases)
         except TypeError:
             raise TypeError("The `bases` option must be a list or a tuple.")
 
@@ -311,3 +311,46 @@ class HistoricalObjectDescriptor(object):
         values = (getattr(instance, f.attname)
                   for f in self.model._meta.fields)
         return self.model(*values)
+
+
+class HistoricalChanges(models.Model):
+
+    class Meta:
+        abstract = True
+
+    def __sub__(self, old_history):
+        if not isinstance(old_history, type(self)):
+            raise TypeError(("unsupported operand type(s) for -: "
+                             "'{}' and '{}'").format(
+                            type(self),
+                            type(old_history)))
+        model_delta_class = getattr(self, 'model_delta_class', ModelDelta)
+        changes = {}
+        changed_fields = []
+        for field in self._meta.fields:
+            if hasattr(self.instance, field.name) and \
+               hasattr(old_history.instance, field.name):
+                old_value = getattr(old_history, field.name, '')
+                new_value = getattr(self, field.name)
+                if old_value != new_value:
+                    changes[field.name] = {'old_value': old_value,
+                                           'new_value': new_value}
+                    changed_fields.append(field.name)
+        return model_delta_class(changes,
+                                 changed_fields=changed_fields,
+                                 date=self.history_date,
+                                 user=self.history_user,
+                                 old_history=old_history,
+                                 new_history=self)
+
+
+class ModelDelta:
+
+    def __init__(self, changes, changed_fields, date, user, old_history,
+                 new_history):
+        self.changes = changes
+        self.changed_fields = changed_fields
+        self.date = date
+        self.user = user
+        self.old_history = old_history
+        self.new_history = new_history
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -21,6 +21,7 @@ Authors
 - Jonathan Sanchez
 - Josh Fyne
 - Klaas van Schelven
+- Leticia Portella
 - Martin Bachwerk
 - Marty Alchin
 - Mauricio de Abreu Antunes
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -213,3 +213,22 @@ You can use the ``table_name`` parameter with both ``HistoricalRecords()`` or
         pub_date = models.DateTimeField('date published')
 
     register(Question, table_name='polls_question_history')
+
+
+History Subtraction
+-------------------
+
+When you have two instances of the same HistoricalRecord, such as HistoricalPoll example above,
+you can make subtractions. This will result in a ModelDelta containing some properties such as a
+dictionary with each field changed between this historical records, a list with all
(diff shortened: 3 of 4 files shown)


simple_history/models.py
```diff
@@ -311,3 +311,46 @@ def __get__(self, instance, owner):
...
+                new_value = getattr(self, field.name)
+                if old_value != new_value:
+                    changes[field.name] = {'old_value': old_value,
+                                           'new_value': new_value}
+                    changed_fields.append(field.name)
+        return model_delta_class(changes,
+                                 changed_fields=changed_fields,
+                                 date=self.history_date,
+                                 user=self.history_user,
+                                 old_history=old_history,
```
treyhunner: I'm not sure what we should call these (`old_history`/`new_history`, `old_record`/`new_record`, `old`/`new`, etc.). I'd prefer to be consistent, but I'm not sure whether we've made variable names referring to these objects yet.

simple_history/models.py
```diff
@@ -311,3 +311,46 @@ def __get__(self, instance, owner):
         values = (getattr(instance, f.attname)
                   for f in self.model._meta.fields)
         return self.model(*values)
+
+
+class HistoricalChanges(models.Model):
```
leportella: Nice! I think this will be better as well

simple_history/models.py
```diff
@@ -311,3 +311,46 @@ def __get__(self, instance, owner):
...
+               hasattr(old_history.instance, field.name):
+                old_value = getattr(old_history, field.name, '')
+                new_value = getattr(self, field.name)
+                if old_value != new_value:
+                    changes[field.name] = {'old_value': old_value,
+                                           'new_value': new_value}
+                    changed_fields.append(field.name)
+        return model_delta_class(changes,
+                                 changed_fields=changed_fields,
+                                 date=self.history_date,
```
leportella: Well, on the case I applied here in our company we used this subtractions to build timelines. Since the history is stored as soon as the model changes, I added the user and date from the last historical record. But indeed this was a very specific case. Do you think we should add both?

simple_history/models.py
```diff
@@ -311,3 +311,46 @@ def __get__(self, instance, owner):
...
+                                 user=self.history_user,
```
leportella: Well, as I said on the other comments, we used this to build timelines, so the user and date that the instance changed was very relevant. Maybe we should keep both.

simple_history/models.py
leportella: I prefer `old_record/newrecord` because we are considering each instance and just `history` is not detailed enough indeed. I think that just  `old`/`new` can be quite confusing.

simple_history/models.py
```diff
@@ -311,3 +311,46 @@ def __get__(self, instance, owner):
...
+        model_delta_class = getattr(self, 'model_delta_class', ModelDelta)
+        changes = {}
+        changed_fields = []
+        for field in self._meta.fields:
+            if hasattr(self.instance, field.name) and \
+               hasattr(old_history.instance, field.name):
+                old_value = getattr(old_history, field.name, '')
+                new_value = getattr(self, field.name)
+                if old_value != new_value:
+                    changes[field.name] = {'old_value': old_value,
```
leportella: Don't know what nametuple are. I will look into this and get back at you!

simple_history/models.py
treyhunner: Maybe we should add neither for now since `new_history` and `old_history` allow fetching each individually.

We can always add them later if we decide there's a good use case for copying them over to make them more easily accessible.

simple_history/models.py
treyhunner: That makes sense to me. Let's go with `old_record` and `new_record` then.

simple_history/models.py
treyhunner: I think maybe we should do neither here as well. If it becomes burdensome to type `my_delta.new_record.user` and `my_delta.old_record.user` we can make shortcuts later.

I'm open to re-considering and adding this and/or the historical date feature, but I'd like to not add them unless we're sure they're particularly helpful. One less feature to maintain? 😄

simple_history/models.py
treyhunner: Great!

`namedtuple` allows making a tuple-like class on the fly, like this:

```python
>>> from collections import namedtuple
>>> HistoricalChange = namedtuple('HistoricalChange', ['old', 'new'])
>>> change = HistoricalChange(old=5, new=6)
>>> change.old
5
>>> change.new
6
>>> old, new = change
>>> old
5
>>> new
6
```

simple_history/models.py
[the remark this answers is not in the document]
merwok: What about returning a timedelta?

johnny-die-tulpe:
sadly no progress since long.. Can we still expect a merge?

rossmechanic:
@johnny-die-tulpe Going to be working on getting this merged in the next few months

simple_history/models.py
```diff
@@ -311,3 +311,46 @@ def __get__(self, instance, owner):
...
                   for f in self.model._meta.fields)
         return self.model(*values)
+
+
+class HistoricalChanges(models.Model):
+
+    class Meta:
+        abstract = True
+
+    def __sub__(self, old_history):
```
kseever: My C++ background left me with a bad taste in my mouth for operator overloads. I tend to favor explicit method calls.
```python
delta = current.changesSince(prior)
```

atodorov:
@kseever, @rossmechanic any update on this ? It looks like this is exactly what I need so is there anything I can do to help/speed-up the inclusion of this PR ?

rossmechanic:
Closing this with #416

PR #283, opened 2017-06-06
Polish translations update

grzegorzbialy:
Hi,

Thanks for calling :) Polish translations updated as requested in PR #281.


treyhunner:
Thanks @grzegorzbialy. It looks like `simple_history/` is removed from the beginning of many lines in this change. I see #281 includes `simple_history/` at the beginning of each line.

I'm not well-informed about the translation files. Which way is the correct one?

rossmechanic:
Verified that the comment string with path doesn't matter.

PR #296, opened 2017-06-21
Fix/issue293

### LINKED ISSUE #293: Documentation item "save without historical record" doesn't render properly
The [recent addition to the documentation](https://github.com/treyhunner/django-simple-history/commit/081c04a3491615acf883cd2e3e059e00fa504eee), doesn't render properly.
(The sentence "You want save model without saving a historical record. You can use this method." also doesn't read well. Perhaps "If you want to save a model without a historical record, you can use the following:".)

evertrol:
Fixes #293 

Notes: 

- I have not verified the actual code, but I would recommend someone double-checks this: the original example code had the method definition `def save_without_historical_record(self, *args, **kwargs):` *outside* the class definition, and was thus invalid. It also contained a few typos, further invalidating the code sample.

- I have not updated the changes file: it doesn't look like documentation changes, or simply minor fixes, are kept in there.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -280,31 +280,25 @@ Or you can pass the change reason explicitly:
     poll.save()
     update_change_reason(poll, 'Add a question')
 
-Save Without Historical Record
-------------------------------
+Save without a historical record
+--------------------------------
 
-You want save model without saving a historical record. You can use this method. 
-
-.. code-block:: python
-    def save_without_historical_record(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


macro1:
Looks good. Please feel free to add yourself to AUTHORS.rst. I agree with the current convention of keeping documentation changes out of CHANGES.rst, so no need to make any changes there.

evertrol:
I'm fine as is; it's just a minor documentation fix, nothing original I came up with myself.
Please feel free to merge as is.

PR #301, opened 2017-07-11
Switch ugettext to ugettext_lazy in models.py

### LINKED ISSUE #288: Change ugettext to ugettext_lazy?
I am using Django signals. This involves putting some imports into the app __init__ file. 

After adding django-simple-history to one of my models, I get this error when starting the dev server:

`django.core.exceptions.AppRegistryNotReady: The translation infrastructure cannot be initialized before the apps registry is ready. Check that you don't make non-lazy gettext calls at import time.`

I believe this error is caused by the ugettext calls in simple_history.models. Is there a reason you chose to use ugettext instead of ugettext_lazy?

haikuginger:
Fixes #288.

We have a use case where we need to import this package before the app registry is ready. This pull request changes `ugettext` calls in `models.py` to `ugettext_lazy`.

If we can get this in a 1.9.1 release soon, that'd be great, but if not, then we can just point at the master hash once it's merged.

@itsjeyd, can you check this code for me?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -38,6 +38,7 @@ Authors
 - Lucas Wiman
 - Michael England
 - Gregory Bataille
+- Jesse Shapiro
 
 Background
 ==========
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Changes
 =======
 
+Unreleased
+----------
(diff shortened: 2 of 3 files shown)


itsjeyd:
@haikuginger The code LGTM. I'm currently building a new devstack for testing the changes; will formally approve once I've verified the changes according to the instructions you left on the internal ticket.

PR #303, opened 2017-07-15
Use get_queryset rather than model.objects

WnP:
This able user to overwrite `get_queryset` safely.

For example I've got a meta model for soft delete, with a boolean `is_void`, which is by default filtered to `False` by meta model manager `get_queryset` method, note that this manager defined also a method `all_include_void`, so in the model admin I can use this latest method in model admin `get_queryset`, but without this fix it just doesn't work as expected.

let me know your thoughts.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -32,6 +32,7 @@ Authors
 - Rod Xavier Bondoc
 - Ross Lote
 - Steven Klass
+- Steeve Chailloux
 - Trey Hunner
 - Ulysses Vilela
 - vnagendra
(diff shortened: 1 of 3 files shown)


macro1:
@WnP List yourself in [AUTHORS.rst](https://github.com/treyhunner/django-simple-history/blob/master/AUTHORS.rst), add your change to [CHANGES.rst](https://github.com/treyhunner/django-simple-history/blob/master/CHANGES.rst), and we'll get this merged!

WnP:
@macro1 done, thanks for your quick answer and all your works on this project.

macro1:
@WnP I've definitely been slacking lately. But thanks for that :)

PR #306, opened 2017-07-18
fix mongodb engine check

singlerider:
models.py line 315
`if settings.DATABASES[connection]['ENGINE'] in ('django_mongodb_engine',):`
can throw a KeyError if the project configuration does not follow
Django docs guidelines for setup

Legacy projects can potentially have a working Django app set up like this:
```python
DATABASES = {
    'default': dj_config_url.config(env='DATABASE_URL'),
    'alternative': dj_config_url.config(env='ALTERNATIVE_DATABASE_URL'),
}
```

Considering that during a migration for simple_history, models.py
checks for a key that does not exist, an exception is thrown that
prevents the project from running.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -40,6 +40,7 @@ Authors
 - Michael England
 - Gregory Bataille
 - Jesse Shapiro
+- Shane Engelman
 
 Background
 ==========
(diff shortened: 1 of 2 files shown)


macro1:
Still an awkward check, but I am okay with this if it prevents some issues.

macro1:
@singlerider would you like to add yourself to the contributors file before I merge this?

singlerider:
@macro1 Thanks. I added myself to AUTHORS. 👍

Issue #177, opened 2015-06-17
Add note in docs for creating history migrations


disdanes:
The docs for simple history are pretty good! I think something that would help user's of your package is to add a line in the docs that after you've registered your history models you'll need to create a migration in order to start using history.

That might be obvious once people go to use it and run into missing table errors, but I think making it explicit in the documentation will help alleviate any 'what do I do now to make it work?' moments.

treyhunner:
@disdanes good point.  Thanks for submitting this issue!

The [documentation files are here](https://github.com/treyhunner/django-simple-history/tree/master/docs) for anyone who sees this issue and wants to take a stab at creating a pull request.

claytondaley:
issue needs closed by someone with rights (resolved by #314)

PR #314, opened 2017-09-24
Add note about running migrations

rca:
I ran into #177 myself, and scratched my head for way too much time (long story short: I was using a variant of this project and to me it appeared that some fields were being dropped when really it was a missing migration :/).

Here's my update to the docs; short and sweet.  Please let me know if more is needed.

Thanks!


macro1:
Thanks @rca. I'm sure we will still get confused devs stumbling into our issue tracker, but hopefully this will ease the process for most.

macro1:
@rca would you like to add yourself to the contributors before I merge this?

PR #318, opened 2017-10-14
- Add ability to see all change history of model.

Mikelarg:


rossmechanic:
You should be able to access the whole change history of a model through the admin by registering the historical model with the admin. I.e. if you have a model called `City` and you want to be able to see all change history for `City`, you can register `HistoricalCity` in the admin. Not sure why you'd want this extra endpoint when you can easily do that.

PR #319, opened 2017-10-17
Self referencing field migration fix

KevinGrahamFoster:
This PR attempts to fix https://github.com/treyhunner/django-simple-history/issues/278 

I am not sure why the existing tests work, but a self referencing key fails for me in practice since the Historical Model attempts to have the self referenced field be of type HistoricalObj instead of just Obj. This PR fixes that.  

Here is the situation. I have the following model

```
class Node(models.Model):
    id = UUIDField(primary_key=True, editable=False, default=uuid.uuid4)
    parent = models.ForeignKey(Node, blank=True, null=True
    history = HistoricalRecords()
```

The migration for this model is generated as follows:

```
            migrations.CreateModel(
            name='HistoricalNode',
            fields=[
                ('id', models.UUIDField(db_index=True, default=uuid.uuid4, editable=False)),
                ('parent', models.ForeignKey(blank=True, db_constraint=False, null=True, on_delete=django.db.models.deletion.DO_NOTHING, related_name='+', to='example_project.HistoricalNode')),
            ],
...
        ),
```

Note that the parent is set to `example_project.HistoricalNode` which is incorrect, it should be to `example_project.Node`.

In this PR I check for the specific case that has been breaking for me and directly set it to the correct value.

An improvement is to add a test that addresses this exact case since it seems that current tests do not, however I am not sure how they do not yet and would appreciate discussion on that.


KevinGrahamFoster:
Comment updated to better describe the situation I am having and what the fix does.

macro1:
Nice catch! Obviously an explicit test for this case would be ideal, to make sure it's not missed in the future. But I'll get it merged for now.

KevinGrahamFoster:
Awesome! thanks! Love simple history!!

PR #323, opened 2017-10-24
Fixes history detail view for concrete models that instantiate abstract historical models defined in separate apps

### LINKED ISSUE #308: Can't Access the History Detail View of /app/model/<pk>/history/<history_id>
We have this setup (Django **1.8.18**, SimpleHistory **1.8.2**):
```
# core/models.py
class VersionedAbstractModel(models.Model):
    class Meta:
        abstract = True

# accounts/models.py
class Brand(VersionedAbstractModel):
    ...

```

So, when I run the `makemigrations`, SimpleHistory migration files will be created in **Accounts** (contains the VersionedAbstractModel child models) and **Core** apps (contains the VersionedAbstractModel parent model). Everything is working fine except when going to the Detail View of particular history objects (e.g. `/accounts/brand/4/history/3/`) since the breadcrumbs in `object_history_form.html` are not resolving properly. It will reference the `/core` admin URL for the Core app (which is non-existent in our case).

Moreover, breadcrumb will switch from `Home › Accounts › Brands › Brand-1 › History` to `Home › Core › Historical Brands › Brand-1 › History › Revert Brand` as you go from history list to history details/revert pages. Hence, it switches scope from **Accounts** app to **Core** app. Unfortunately, we have **NO** non-abstract model registered in our **Core** app and it only contains services/utility/helper functions, plus the migration files for every model due to the inherited VersionedAbstractModel. This issue probably happens when `inherit=True` and/or the involved `parent-child` models are in separate apps/namespaces.

Error Trace:
```
NoReverseMatch at /accounts/brand/4/history/3/
Reverse for 'app_list' with arguments '(u'core',)' and keyword arguments '{}' not found. 
1 pattern(s) tried: [u'(?P<app_label>accounts|events)/$']

In template ~/.virtualenvs/eventcms/eventcms/templates/simple_history/object_history_form.html,
error at line 8:
<a href="{% url "admin:app_list" app_label %}">{{app_label\|capfirst\|escape}}</a> &rsaquo;
```

Possible Workarounds:
1. Create a dummy model in our **Core** app and register it via `admin.site.register()` so that `/core` URL will be registered. However, the dummy model will show in Admin dashboard.
2. Override the `object_history_form.html` and remove the line referencing the **Core** app or the app containing the parent `SimpleHistory` model (i.e. the `VersionedAbstractModel`).

Eventually, we did the item 2 since we have no need for the `/core` URL as of now. It would be great/interesting if you could find a cleaner/simpler solution.

Cheers for this great app! :)

rossmechanic:
Fixes #308 .

### DIFF
--- a/simple_history/admin.py
+++ b/simple_history/admin.py
@@ -162,7 +162,7 @@ class SimpleHistoryAdmin(admin.ModelAdmin):
             'is_popup': False,
             'media': mark_safe(self.media + admin_form.media),
             'errors': helpers.AdminErrorList(form, formsets),
-            'app_label': model._meta.app_label,
+            'app_label': original_opts.app_label,
             'original_opts': original_opts,
             'changelist_url': reverse('%s:%s_%s_changelist' % url_triplet),
             'change_url': reverse('%s:%s_%s_change' % url_triplet,
--- a/simple_history/tests/models.py
+++ b/simple_history/tests/models.py
@@ -342,7 +342,7 @@ class UntrackedConcreteBase(models.Model):
     pass
 
 
-class TrackedConcreteBaseExternal(AbstractExternal):
+class ConcreteExternal(AbstractExternal):
     name = models.CharField(max_length=50)
 
     class Meta:
--- a/simple_history/tests/models.py
+++ b/simple_history/tests/models.py
@@ -1,12 +1,30 @@
 from __future__ import unicode_literals
 
 from django.db import models
+from django import VERSION
 
 from simple_history.models import HistoricalRecords
 from simple_history import register
 
 from .custom_user.models import CustomUser as User
-from .external.models.model2 import AbstractExternal
+
+try:
+    from django.apps import apps
+except ImportError:  # Django < 1.7
+    from django.db.models import get_model
+else:
+    get_model = apps.get_model
+
+# 1.6 has different way of importing models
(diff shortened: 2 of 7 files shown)


simple_history/tests/models.py
```diff
@@ -1,12 +1,31 @@
...
+except ImportError:  # Django < 1.7
+    from django.db.models import get_model
+else:
+    get_model = apps.get_model
+
+# 1.6 has different way of importing models
+if VERSION[:3] >= (1, 7, 0):
+    from .external.models.model1 import AbstractExternal
+else:
+    class AbstractExternal(models.Model):
```
macro1: Not a huge fan of this, but it's 1.6 so... ¯\\_(ツ)_/¯

simple_history/admin.py
```diff
@@ -162,7 +162,7 @@ def history_form_view(self, request, object_id, version_id):
             'is_popup': False,
             'media': mark_safe(self.media + admin_form.media),
             'errors': helpers.AdminErrorList(form, formsets),
-            'app_label': model._meta.app_label,
+            'app_label': original_opts.app_label,
```
macro1: I guess this makes sense. Even if the admin is for the history... From the perspective of the admin the app should line up with the original model.

PR #320, opened 2017-10-20
Prepare for Django 2.0

flesser:
I was trying to get `django-simple-history` working with Django 2.0.
Here are the changes I made to get it run. (No guarantee for completeness or correct downward compatibility)

Issue #291, opened 2017-06-16
Address warnings from Django 1.11


merwok:
These are the warnings when running with Django 1.11:
* Importing from django.core.urlresolvers is deprecated in favor of django.urls.
* Usage of ForeignObjectRel.to attribute has been deprecated. Use the model attribute instead.
* Usage of field.rel has been deprecated. Use field.remote_field instead.
* django.utils.translate.string_concat() is deprecated in favor of django.utils.text.format_lazy().
* permalink() is deprecated in favor of calling django.urls.reverse() in the decorated method.

I could see how simple_history handles compat and make a PR, but not until a few weeks.

seocam:
Also
* Using user.is_authenticated() and user.is_anonymous() as a method is deprecated. Remove the parentheses to use it as an attribute.

merwok:
PR #320 addresses some of the issues.

merwok:
Great news!

PR #331, opened 2017-12-06
Handle reverse move and models.permalink removal

jmbowman:
This should handle two of the changes in recent Django versions mentioned in #291 -

* The move of `reverse()` from `django.core.urlresovers` to `django.urls`
* The removal of `models.permalink`

Compatibility with all currently supported Django versions should be maintained.  I realize that #320 also has a fix for the first point, but it seems to have gotten bogged down in figuring out how to handle some of the other changes; it might make sense to just merge these simple fixes first.

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -21,6 +21,10 @@ try:
 except ImportError:  # Django < 1.7
     from django.db.models import get_app
 try:
+    from django.urls import reverse
+except ImportError:  # Django < 1.10
+    from django.core.urlresolvers import reverse
+try:
     from south.modelsinspector import add_introspection_rules
 except ImportError:  # south not present
     pass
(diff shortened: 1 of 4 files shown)


simple_history/models.py
```diff
@@ -204,14 +208,12 @@ def get_extra_fields(self, model, fields):
 
         user_model = getattr(settings, 'AUTH_USER_MODEL', 'auth.User')
 
-        @models.permalink
         def revert_url(self):
```
rossmechanic: Are there any tests in the test suite that are testing this function? Couldn't find any. Want to add one and include it with the PR?

jmbowman: @rossmechanic I added a simple test of the `reverse_url()` output; there was an existing test that called it, but this looks like the first one to sanity check the output.

rossmechanic:
Fixed in #363

Issue #174, opened 2015-05-09
Update documentation about how simple history does not log history when .update() or bulk_create are used


vinay-kotian:
When .update() function is used to update a query set result the history of the values changed is not created in the history table. Please can you look into this issue, let me know if any help is required from my end. 
Thank You 
- Vinay Kotian

macro1:
We use the `post_save` signal to trigger new historical records being created. That signal is not fired when `update()` is used instead of `save()`. If you wish to use `update()` for performance reasons, you could create the historical records manually using the `Model.history` manager.

From the [`update()` documentation][https://docs.djangoproject.com/en/dev/ref/models/querysets/#update]:

> ...realize that `update()` does an update at the SQL level and, thus, does not call any `save()` methods on your models, nor does it emit the `pre_save` or `post_save` signals (which are a consequence of calling `Model.save()`). If you want to update a bunch of records for a model that has a custom `save()` method, loop over them and call `save()`...

Skyross:
Hi guys!
Thank you for good package. I used this workaround for my needs. This is example:

```python
from django.utils.timezone import now

class DeletableModelQuerySet(models.QuerySet):
    def update(self, **kwargs):
        if hasattr(self.model, 'history'):
            cache, model = [], None

            for instance in self.all():
                if not model:
                    model = instance.history.model

                attrs = {}
                for field in instance._meta.fields:
                    attrs[field.attname] = getattr(instance, field.attname)

                cache.append(model(history_date=now(), history_type='~', history_user=None, **attrs))

            if model and cache:
                model.objects.bulk_create(cache)

        return super(DeletableModelQuerySet, self).update(**kwargs)

class LogicDeleteManager(models.Manager):
    """
    Model Manager for filtering logical deleted objects
    """
    def get_queryset(self):
        return DeletableModelQuerySet(self.model, using=self._db).filter(deleted=False)

class DeletableModel(models.Model):
    """
    Common model with `soft` deleting. All deletable models should inherit this as base model.
    For `hardware` (real) deletion use objects_with_deleted Model Manager.
    """
    deleted = models.BooleanField(default=False)

    objects = LogicDeleteManager()
    objects_with_deleted = models.Manager()

    class Meta:
        abstract = True

    def delete(self, using=None):
        self.deleted = True
        self.save()

    def real_delete(self, using=None):
        super(DeletableModel, self).delete(using)
```

jpatel3:
[...]

PR #335, opened 2017-12-22
fix out-of-memory exceptions while generating historical models with …

markddavidoff:
…management command

### DIFF
--- a/simple_history/management/commands/_populate_utils.py
+++ b/simple_history/management/commands/_populate_utils.py
@@ -1,3 +1,4 @@
+from django import db
 from django.utils.timezone import now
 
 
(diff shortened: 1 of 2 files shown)


markddavidoff:
I had too many model instances, this fixed it for me.

simple_history/management/commands/_populate_utils.py
```diff
@@ -15,15 +16,45 @@ def get_history_model_for_model(model):
...
+    historical_instances = []
+    count = 1
+    chunk_count = 0
+    for instance in model.objects.iterator():
+        if count > max_in_memory_models:
+            count = 0
+            start_num = chunk_count * max_in_memory_models
+            end_num = start_num + max_in_memory_models
+            stdout.write(
+                "Starting bulk creating history models for {} instances {}-{}"
```
markddavidoff: maybe delete or hide behind a flag these prints its a lot of logging if its low and you have a lot of models. The batch doesn't print.

rossmechanic:
Opening an issue for this describing the issue (#398) you ran into, and then we'll put a PR up.

rossmechanic:
@markddavidoff Mind fixing the errors and merging conflicts and then I'll review?

markddavidoff:
@treyhunner Removed that extra arg I added and added comments.

rossmechanic:
@markddavidoff looks like this is failing flake8 if you want to fix those linting issues.

markddavidoff:
T_T

simple_history/management/commands/_populate_utils.py
```diff
@@ -15,16 +16,57 @@ def get_history_model_for_model(model):
...
-                field.attname: getattr(instance, field.attname)
-                for field in instance._meta.fields
-            }
-        ) for instance in model.objects.all()]
-    history_model.objects.bulk_create(historical_instances,
-                                      batch_size=batch_size)
+def bulk_history_create(
+        log_file, model, history_model, batch_size):
+    """Save a copy of all instances to the historical model.
+    :param log_file: file object used to log to, can pass sys.stdout for
```
rossmechanic: Why the addition of a log_file? Don't love changing this API to include this. Don't think we log anywhere else. Also planning to use this outside of just management commands so I'd avoid this logging here.

rossmechanic:
@markddavidoff Mind working on the test coverage for this?

simple_history/management/commands/_populate_utils.py
```diff
@@ -15,16 +16,57 @@ def get_history_model_for_model(model):
...
-            history_date=getattr(instance, '_history_date', now()),
-            history_user=getattr(instance, '_history_user', None),
-            **{
-                field.attname: getattr(instance, field.attname)
-                for field in instance._meta.fields
-            }
-        ) for instance in model.objects.all()]
-    history_model.objects.bulk_create(historical_instances,
-                                      batch_size=batch_size)
+def bulk_history_create(
```
rossmechanic: This isn't transaction-safe. I don't love the idea of wrapping this in a transaction either, since it may block for an unacceptable amount of time depending on table size. Maybe add an option to wrap this in a transaction? Or at least document that this isn't transaction-safe.

markddavidoff: why is it now not transaction safe?

simple_history/management/commands/_populate_utils.py
markddavidoff: in management commands you traditionally log to sys.stdout and I'd prefer to have logging here that works in the management command. would logger.info still show up in the management command?

markddavidoff:
haven't found time to add tests for this yet

simple_history/management/commands/_populate_utils.py
rossmechanic: The `bulk_create` itself is wrapped in an atomic transaction, but you're calling a series of `bulk_create` calls that aren't all together wrapped in a transaction. So if one of those bulk_create calls fails, then you'll have some of your data inserted in the database, rather than rolling it back so that none of your data is inserted.

rossmechanic:
@markddavidoff happy to take this over and add to the coverage if you don't have time. Looking to get this merged ASAP because it's relevant to some of the bulk_create work I want to do for #380 and #174.

simple_history/management/commands/_populate_utils.py
markddavidoff: gotcha. thanks.

markddavidoff:
feel free to take over, thanks.

simple_history/management/commands/_populate_utils.py
rossmechanic: Fine with making logging an optional parameter here (which can be triggered by a management command flag). Does that work for you?

rossmechanic:
Closed with #408

PR #340, opened 2018-01-12
update for django 2.0

Czaki:
django.core.urlresolvers is changed to django.urls


Czaki:
all changes are in django-20 branch

Issue #116, opened 2014-09-16
Override history ID?


estebistec:
Is it possible to override the history_id of a history model, much in the same way you document overriding the history date?

I'm going to need to expose the history over URLs and I'd rather not expose incrementing auto-integers in URLs.

rossmechanic:
Fixed in #368

PR #351, opened 2018-03-15
Add new constructor options to HistoricalRecords class

rwlogel:
- Provide history_id_field to change the field type for history_id
- Provide user_model to use an alternative model for tracking history_user

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -44,6 +44,7 @@ Authors
 - Jesse Shapiro
 - Kevin Foster
 - Shane Engelman
+- Ray Logel
 
 Background
 ==========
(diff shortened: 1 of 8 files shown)


rossmechanic:
This would solve for #116. Happy to discuss allowing UUIDs for the historical tables, but I'm less inclined to allow for a separate user model. Want to split this up into two PRs and we can discuss them separately? @rwlogel

rossmechanic:
No worries. Let's definitely split the changes into two PRs so that we can have the discussions separately. Happy to allow UUIDs on history IDs as I've seen that functionality requested multiple times. 

More hesitant about allowing a CustomUser model as it's rare to see Django programmers using a user model that doesn't extend `AUTH_USER_MODEL`, and I'm worried about any possible side effects (the way we track users with middleware is a little magic-y). 

However, I understand your use case – essentially keeping sensitive fields in a separate table, right? Going to close this PR and ask you to open these two features in separate PRs and we can discuss them separately.

rwlogel:
Changing the user model doesn't work with the built in middleware, as documented in the PR it requires the user of the `_history_user`.  I will submit new PR's shortly and try to include an example of what a middleware with the a custom user model may look like.

rossmechanic:
Thanks @rwlogel

PR #324, opened 2017-11-07
Fixes issue with tracking an inherited model

rossmechanic:
Fixes #269 . @macro1 I wrote this so that it handles the different use cases (when `HistoricalRecords` is set as a field vs. when `register` is used), but am open to another solution if you have a cleaner one that comes to mind. For an abstract base model defined in app1 and a concrete implementation defined in app2, this makes sure that the historical model is also in app2 as opposed to app1.

PR #352, opened 2018-03-16
Correct module for history model of inherited model

rwlogel:
If base model is in a different module the history model for the
inherited model should be in the same module as the inherited model

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -44,6 +44,7 @@ Authors
 - Jesse Shapiro
 - Kevin Foster
 - Shane Engelman
+- Ray Logel
 
 Background
 ==========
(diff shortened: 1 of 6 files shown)


rwlogel:
The coverage tests are failing because the error conditions of the unit tests are not being executed which is the expected result.  Should these tests be in a different file or implemented differently?

rwlogel:
One of the key features missing from #324 is the change to how `importlib.import_module` (models.py lines 96-101 of this diff) is called.  I think that the code should actually be using `history_model.__module__` all the time but I was concerned with breaking more things then I had time to test for.

rwlogel:
For example if you add the unit tests added to this request to the #324 request they both fail because the `ConcreteExternal` and `ConcreteExternal2` class are not registered in the `tests.models`  module.

rossmechanic:
Wanted to add some tests and change a few issues with wording so opened a new PR with that that I'm about to merge (#355). Thanks for the solution @rwlogel

Issue #344, opened 2018-01-30
History of custom UserModel 


Czaki:
I try to add to my app history tracking of custom model. To get it working I need to comment history field, then make migrations, then uncomment and again create migrations. in other case Its create History model before User model and there is error in migration. 
Do you have any suggestion how to avoid it? **Or how to debug it?**
 In other models from this app I have split on creating main part of model and Add field with reference. 

Django 2.0, python 3.6, version from branch django-20

rossmechanic:
Maybe check the order in which your apps are imported? Ideally, if you could include the minimum version of your app that includes this bug, I could help with debugging it.

PR #353, opened 2018-03-18
Add support for Django 2.0

macro1:
Add support for Django 2.0. I am not running this in production, and I've already seen one issue raised with concerns around migrations (#344), so additional testing anyone can offer is extremely valuable.

I will be going back through the comments on the other pr to be sure I'm not missing anything brought up in #320.

closes #320 (This branch is based on flesser's.)


macro1:
It looks like we can keep using #320 with my changes.

PR #363, opened 2018-04-03
More Django 1.9 leftovers

wizpig64:
Follow up to #362 

- Removed a pointless MiddlewareMixin alias in simple_history/middleware.py that was part of an old compatibility shim.
- Tests now use the MIDDLEWARE setting rather than MIDDLEWARE_CLASSES. This isn't required until Django 2.0, but now that 1.9 is dropped I see no reason not to switch over now.
- Import reverse() from django.urls rather than deprecated django.core.urlresolvers module.
- Removed usage of django's deprecated @models.permalink decorator. (Deprecated in 1.11, but the fix might as well be done now.)

Thanks :)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -34,6 +34,7 @@ Authors
 - Marty Alchin
 - Mauricio de Abreu Antunes
 - Micah Denbraver
+- Phillip Marshall
 - Rajesh Pappula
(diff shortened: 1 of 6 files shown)


rossmechanic:
lgtm. Thanks for the changes @wizpig64

PR #365, opened 2018-04-05
feat(HistoryRecords): prev and next attributes

### LINKED ISSUE #230: Query previous/next historical entry
One of the main use cases I've found with django-simple-history has been to get the previous historical entry to calculate deltas in values.

Issue #201 talks about the Django utility functions, which allows for getting the last value by a field. I think it would be great to add a `previous` property to the historical model that grabbed the `get_previous_by_history_date` that reflected the same tracked instance automatically. I feel like a `next` and `previous` method fits perfectly within the historical records context.

As a dependent though, comparing versions between points in time would be excellent, too, but I'll submit a second issue based on response to this.

SpainTrain:
TODO
----

- [x] Once API is approved, add docs on how to use this feature

Changes
-------

- New attributes on history model retrieve previous and next records

Closes #230

---
This change is [Reviewable](https://reviewable.io/reviews/treyhunner/django-simple-history/365)

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -9,6 +9,7 @@ from django.apps import apps
 from django.conf import settings
 from django.contrib import admin
 from django.db import models, router
+from django.db.models import Q
 from django.db.models.fields.proxy import OrderWrt
 from django.urls import reverse
 from django.utils import six
@@ -222,6 +223,22 @@ class HistoricalRecords(object):
                 for field in fields.values()
             })
 
+        def get_next_record(self):
+            """
+            Get the next history record for the instance. `None` if last.
+            """
+            return self.instance.history.filter(
+                Q(history_date__gt=self.history_date)
+            ).order_by('history_date').first()
+
+        def get_prev_record(self):
+            """
+            Get the previous history record for the instance. `None` if first.
+            """
+            return self.instance.history.filter(
+                Q(history_date__lt=self.history_date)
+            ).order_by('history_date').last()
+
         if self.history_id_field:
             history_id_field = self.history_id_field
             history_id_field.primary_key = True
@@ -252,6 +269,8 @@ class HistoricalRecords(object):
             ),
             'instance': property(get_instance),
             'instance_type': model,
+            'next_record': property(get_next_record),
+            'prev_record': property(get_prev_record),
             'revert_url': revert_url,
             '__str__': lambda self: '%s as of %s' % (self.history_object,
                                                      self.history_date)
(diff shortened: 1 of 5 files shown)


simple_history/models.py
```diff
@@ -221,6 +221,32 @@ def get_instance(self):
...
+                entry_hist_idx = hist_list.index(self)
+            except ValueError:
+                return None
+            return hist_list[entry_hist_idx - 1] if entry_hist_idx > 0 else None
+
+        def get_prev_record(self):
+            """
+            Get the previous history record for the instance. `None` if first.
+            """
+            entry = self.instance
```
rossmechanic: I don't love going from historical record to instance back to historical record. Ideally we'd avoid this. Maybe look to see what they do in `cleanerversion`? https://github.com/rossmechanic/cleanerversion/blob/master/versions/models.py#L83

I think something similar to that on our historical tables would be a much cleaner solution

simple_history/models.py
```diff
@@ -222,6 +223,22 @@ def get_instance(self):
                 for field in fields.values()
             })
 
+        def get_next_record(self):
+            """
+            Get the next history record for the instance. `None` if last.
+            """
+            return self.instance.history.filter(
```
SpainTrain: Good feedback.  This forced me to implement these in a much more efficient and clean way.  Thanks!

SpainTrain:
@rossmechanic should be good for another review.  If reasonable, I will add relevant docs.  Thanks!

rossmechanic:
@SpainTrain need to pass the `flake8` linting errors first.

SpainTrain:
derp :white_check_mark:

SpainTrain:
docs added

rossmechanic:
@SpainTrain Thanks. This looks way better to me at first glance. Going to spend some time playing around with it and get you comments by end of week.

rossmechanic:
Mind adding your name to `AUTHORS.rst` and the change to `CHANGES.rst`? Also add the PR number after the change (i.e. (gh-365))

simple_history/models.py
```diff
@@ -222,6 +223,22 @@ def get_instance(self):
...
+                Q(history_date__gt=self.history_date)
+            ).order_by('history_date').first()
+
+        def get_prev_record(self):
+            """
+            Get the previous history record for the instance. `None` if first.
+            """
+            return self.instance.history.filter(
+                Q(history_date__lt=self.history_date)
+            ).order_by('history_date').first()
```
ncvc: This should reverse order by history date or take `.last()` with the existing order. We should also use say 3 records in `test_get_prev_record` / `test_get_next_record` and get the previous and next records of the last / first records, respectively, to ensure we catch this type of bug in the unit tests.

rossmechanic: Good catch @ncvc

SpainTrain: very good catch, thanks!

PR #367, opened 2018-04-06
Removed reference to south package.

wizpig64:
South isn't compatible with Django 1.7 or higher, and we require 1.11. There's no scenario where this try block can do anything but pass. Let's get rid of it.


rossmechanic:
Thanks @wizpig64

PR #368, opened 2018-04-10
Provide history_id_field to change the field type for history_id

rwlogel:
@rossmechanic here is the history id field change by itself.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Changes
 - Dropped support for Django<=1.10 (gh-356)
 - Fix bug where history_view ignored user permissions (gh-361)
 - Fixed HistoryRequestMiddleware which hadn't been working for Django>1.9 (gh-364)
+- Add ability to specify custom history_id field
 
 1.9.1 (2018-03-30)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ Changes
 - Dropped support for Django<=1.10 (gh-356)
 - Fix bug where history_view ignored user permissions (gh-361)
 - Fixed HistoryRequestMiddleware which hadn't been working for Django>1.9 (gh-364)
-- Add ability to specify custom history_id field
+- Add ability to specify custom history_id field (gh-368)
 
 1.9.1 (2018-03-30)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,13 +1,16 @@
 Changes
 =======
 
+Unreleased
+----------
+- Add ability to specify custom history_id field (gh-368)
+
 2.0 (2018-04-05)
 ----------------
 - Added Django 2.0 support (gh-330)
 - Dropped support for Django<=1.10 (gh-356)
 - Fix bug where history_view ignored user permissions (gh-361)
 - Fixed HistoryRequestMiddleware which hadn't been working for Django>1.9 (gh-364)
-- Add ability to specify custom history_id field (gh-368)
 
 1.9.1 (2018-03-30)
 ------------------
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -133,11 +133,11 @@ Custom ``history_id``
 By default, the historical table of a model will use an ``AutoField`` for the table's
 ``history_id`` (the history table's primary key). However, you can specify a different
 type of field for ``history_id`` buy passing a different field to ``history_id_field``
-parameter.  A common use case for this would be to use a ``UUIDField``.
+parameter.
 
-You can also change the default ``history_id`` field for all classes using the
-setting ``SIMPLE_HISTORY_HISTORY_ID_DEFAULT_FIELD``.  This setting can still be overridden
-using the ``history_id_field`` parameter on a per model basis.
+A common use case for this would be to use a ``UUIDField``.  If you want to use a ``UUIDField``
+as the default for all classes set ``SIMPLE_HISTORY_HISTORY_ID_USE_UUID=True`` in the settings.
+This setting can still be overridden using the ``history_id_field`` parameter on a per model basis.
 
 You can use the ``history_id_field`` parameter with both ``HistoricalRecords()`` or
 ``register()`` to change this behavior.
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -130,12 +130,16 @@ Admin integration requires that you use a ``_history_user.setter`` attribute wit
 Custom ``history_id``
 ---------------------
 
-If you want to use something other then an ``AutoField`` for the history
-table primary key you can pass in a ``history_id_field``.  A common use case
-for this would be to use a ``UUIDField``.
+By default, the historical table of a model will use an ``AutoField`` for the table's
+``history_id`` (the history table's primary key). However, you can specify a different
+type of field for ``history_id`` buy passing a different field to ``history_id_field``
+parameter.  A common use case for this would be to use a ``UUIDField``.
 
-For any field provided it will automatically set ``primary_key=True`` and
-``editable=False``.
+You can use the ``history_id_field`` parameter with both ``HistoricalRecords()`` or
+``register()`` to change this behavior.
+
+Note: regardless of what field type you specify as your history_id field, that field will
+automatically set ``primary_key=True`` and ``editable=False``.
 
 .. code-block:: python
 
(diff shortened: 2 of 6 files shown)


docs/advanced.rst
```diff
@@ -127,6 +127,29 @@ referencing the ``changed_by`` field:
...
+.. code-block:: python
+
+    import uuid
+    from django.db import models
+    from simple_history.models import HistoricalRecords
+
+    class Poll(models.Model):
+        question = models.CharField(max_length=200)
+        pub_date = models.DateTimeField('date published')
+        history = HistoricalRecords(
```
rossmechanic: So this only works if I add history to the model using `HistoricalRecords`? You're also going to want to edit the `register` function since that is the other way to give a model history

docs/advanced.rst
```diff
@@ -127,6 +127,29 @@ referencing the ``changed_by`` field:
 
 Admin integration requires that you use a ``_history_user.setter`` attribute with your custom ``_history_user`` property (see :ref:`admin_integration`).
 
+Custom ``history_id``
+---------------------
+
+If you want to use something other then an ``AutoField`` for the history
```
rossmechanic: "By default, the historical table of a model will use an `AutoField` for the table's `history_id` (the history table's primary key). However, you can specify a different type of field for `history_id` but passing a different field to `history_id_field` parameter." Something like that sounds a little more clear to me

docs/advanced.rst
```diff
@@ -127,6 +127,29 @@ referencing the ``changed_by`` field:
...
+table primary key you can pass in a ``history_id_field``.  A common use case
+for this would be to use a ``UUIDField``.
+
+For any field provided it will automatically set ``primary_key=True`` and
+``editable=False``.
```
rossmechanic: "Note: regardless of what field type you specify as your `history_id` field, that field will automatically set ``primary_key=True`` and ``editable=False``."

docs/advanced.rst
```diff
@@ -127,6 +127,29 @@ referencing the ``changed_by`` field:
...
+
+    import uuid
+    from django.db import models
+    from simple_history.models import HistoricalRecords
+
+    class Poll(models.Model):
+        question = models.CharField(max_length=200)
+        pub_date = models.DateTimeField('date published')
+        history = HistoricalRecords(
+            history_id_field=models.UUIDField(default=uuid.uuid4)
```
rossmechanic: Might you just want to add a setting that toggles UUIDField vs AutoField? Or do you think it's important to allow any type of field here?

simple_history/registry_tests/tests.py
```diff
@@ -88,28 +88,28 @@ class TestTrackingInheritance(TestCase):
 
     def test_tracked_abstract_base(self):
         self.assertEqual(
-            [
+            sorted([
```
rossmechanic: Why the `sorted` additions?

rwlogel: Originally I added it because some of the tests were failing because the order of the meta fields changed when I supplied the `history_id`.  I just checked and that doesn't seem to be the case anymore maybe it was only the older versions of django that were supported at the time, but have since been dropped.  I can remove this change.

docs/advanced.rst
rwlogel: It does work with `register`, I will update the docs and unit tests so that is tested.

docs/advanced.rst
rwlogel: We have actually created a model `class AutoUUIDField(models.UUIDField)` which implements the `pre_save` as outlined which is also why I would prefer to keep it as a field instead of a simple flag.

simple_history/registry_tests/tests.py
rossmechanic: Cool. Let's remove this

docs/advanced.rst
rossmechanic: Oh yes, it gets passed through `**records_config`. Please update docs and tests for that

rwlogel:
@rossmechanic I fixed all the review comments and corrected the flake8 errors.

docs/advanced.rst
rossmechanic: What about setting the default field in settings and allowing that to be overwritable? That way, if I want to use UUIDField on every `history_id` without explicitly stating that, I can do so. I.e., if `SIMPLE_HISTORY_HISTORY_ID_DEFAULT_FIELD` is undefined, then use `AutoField` as the default, otherwise use whatever is defined in that setting. Then, on a case-by-case basis, a dev can overwrite the `history_id_field` for a specific model. If you did it with the setting, would you even have the need to overwrite on a case-by-case basis?

rossmechanic:
@rwlogel Thanks for getting to this so quickly. What are your thoughts on setting this in `settings.py` instead of on a model-by-model basis? Or at least setting the default in `settings.py` and making that overwritable on a case-by-case basis? 

Also, have you tested the admin UI with these changes? I think the history and revert URLs might fail with UUIDs since they're not strings. I've seen that issue happen before when switching to UUIDs in other libraries. Can you test that out and let me know what happens?

rwlogel:
Adding a default sounds reasonable, I will leave it in the class in case someone does want to do it on a per model basis.  I will investigate the test for the admin, but I am positive that they work because I'm using the admin history in my teams project and it works fine.

rossmechanic:
@rwlogel Cool. That sounds good to me. Once you add default setting with docs, I'm happy to merge

rossmechanic:
@rwlogel also feel free to add your name to `AUTHORS.rst`

docs/advanced.rst
ncvc: @rwlogel @rossmechanic I think having a boolean setting in `settings.py` such as `SIMPLE_HISTORY_HISTORY_ID_USE_UUID` as well as allowing overriding the ID field directly in the `HistoricalRecords` constructor is a good trade-off. This would make it easy to use UUID fields across a project by default (a common substitute for `AutoField`) while still allowing for more customized IDs for those that need it.

rossmechanic: I like this idea. @rwlogel do you mind doing this instead? Solves for the more commonly used issue (using UUIDs instead of sequential IDs) and allows users like yourself to overwrite them with whatever id you want to use, wherever you want to use it

rwlogel: Using `SIMPLE_HISTORY_HISTORY_ID_DEFAULT_FIELD` accomplishes the same thing as  `SIMPLE_HISTORY_HISTORY_ID_USE_UUID` with more flexibility.  How to set it up for UUID is documented in the advanced section so it should be very easy for someone to just copy that.

rossmechanic: But instantiating the field in the settings file prevents us from ever being able to have custom field names for `history_id`. I.e., if in the future, someone is using `history_id` in their model and thus needs to call this historical table's `history_id` something else, this would prevent that, since there's only one instantiation of the UUIDField in settings. Renaming history columns isn't something we support currently, but it's something that people have asked for that I plan to allow in the future to prevent name clashing. If you look at the example below:
```
field = models.CharField(max_length=50)

class TestModel(models.Model):
    test = field

class TestModel1(models.Model):
    test1 = field
```
both `TestModel` and `TestModel1` will have a `CharField` named `test` rather than `TestModel` being named `test` and `TestModel1` being named `test1`. We'll run into the same issue by instantiating in settings. That's why I'd prefer the setting to be a boolean, since most people's use case will be UUIDField (and we can instantiate a new UUIDField each time), and for users who want something other than UUIDField or AutoField, they can overwrite in the `HistoricalRecords` parameters.

rwlogel: Good point

rwlogel:
@rossmechanic @ncvc Change to `SIMPLE_HISTORY_HISTORY_ID_USE_UUID` is ready.

CHANGES.rst
```diff
@@ -7,6 +7,7 @@ Changes
 - Dropped support for Django<=1.10 (gh-356)
 - Fix bug where history_view ignored user permissions (gh-361)
 - Fixed HistoryRequestMiddleware which hadn't been working for Django>1.9 (gh-364)
+- Add ability to specify custom history_id field (gh-368)
```
rossmechanic: This isn't part of the 2.0 release. So you can add another section above like this:
```
Unreleased
------------
- Add ability to specify custom history_id field (gh-368)
```

rossmechanic:
All set. Merging.

PR #370, opened 2018-04-12
Track history of M2M fields [WIP]

avalanchy:
Yet another attempt of introducing many to many fields support for django-simple-history. 

Today I rebased this with master. Before that I was using this for like two months, but it was rebased with `django-20` branch. 

Not sure when I will continue working on this, but for now I decided to share the code with you to collect some feedback and know if this is at least good direction.

Current solution is backward compatible, it means that m2m fields aren't tracked by default. Programmer needs to define which m2m fields should be tracked and pass them within `HistoricalRecords` field declaration (see api 1),

Such field definition is detected by Django `makemigration` and it that will create new table `Historical<M2mTroughModelName>`.

Under the hood tracking version changes and creating historical entries is based on Django signals.

As this is WIP, the only working way of this feature is with the new decorator. This new decorator is a way to manipulate m2m relation and to track their history but without tracking history of original object (see api 2).

api 1. Defining which m2m fields should be tracked:
```
class Business(Model):
    attachments = models.ManyToManyField(Attachment)
    history = HistoricalRecords(m2m_fields=[attachments])
```

api 2. Saving with new decorator, that prevents from creating historical entry of base object:
```
self.save()
with self.use_last_historical_record():
    self.attachments.clear()
```

Looking forward to see some comments. Thanks!

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -3,6 +3,8 @@ from __future__ import unicode_literals
 import copy
 import importlib
 import threading
+from contextlib import contextmanager
+from functools import partial
 
 from django.apps import apps
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


rossmechanic:
Thanks @avalanchy ! I actually like this approach but would prefer to have discussion around this in an issue rather than in an open PR and then move it to PR when we know exactly what we want. Opening #399 to discuss.

PR #371, opened 2018-04-15
Allow alternative user model for tracking history_user

rwlogel:
@rossmechanic  I improved the ability to set an alternative user model so it will also work with register and middleware as discussed in #351.

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -24,13 +24,20 @@ from .manager import HistoryDescriptor
 registered_models = {}
 
 
+def default_get_user(request, **kwargs):
+    try:
+        return request.user
+    except AttributeError:
+        return None
+
+
 class HistoricalRecords(object):
     thread = threading.local()
 
     def __init__(self, verbose_name=None, bases=(models.Model,),
                  user_related_name='+', table_name=None, inherit=False,
-                 excluded_fields=None,
-                 history_id_field=None, user_model=None,
+                 excluded_fields=None, history_id_field=None,
+                 user_model=None, get_user=default_get_user,
                  history_change_reason_field=None):
         self.user_set_verbose_name = verbose_name
         self.user_related_name = user_related_name
@@ -38,6 +45,7 @@ class HistoricalRecords(object):
         self.inherit = inherit
         self.history_id_field = history_id_field
         self.user_model = user_model
+        self.get_user = get_user
         self.history_change_reason_field = history_change_reason_field
         if excluded_fields is None:
             excluded_fields = []
@@ -75,15 +83,11 @@ class HistoricalRecords(object):
 
     def finalize(self, sender, **kwargs):
         inherited = False
-        try:
-            hint_class = self.cls
-        except AttributeError:  # called via `register`
-            pass
-        else:
-            if hint_class is not sender:  # set in concrete
-                inherited = (self.inherit and issubclass(sender, hint_class))
-                if not inherited:
-                    return  # set in abstract
+        if self.cls is not sender:  # set in concrete
+            inherited = (self.inherit and issubclass(sender, self.cls))
+            if not inherited:
+                return  # set in abstract
+
         if hasattr(sender._meta, 'simple_history_manager_attribute'):
             raise exceptions.MultipleRegistrationsError(
                 '{}.{} registered multiple times for history tracking.'.format(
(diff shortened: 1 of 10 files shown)


rossmechanic:
@rwlogel sorry for the delay on this. It touches a lot of the codebase and I want to go through it thoroughly before merging. Hopefully will have it looked at by EOW

rossmechanic:
@rwlogel checking in on this. Is this still something you want to merge?

rwlogel:
@rossmechanic definitely, we are currently using a forked repo for our released product.  It would be great if we could use the version from pip

rossmechanic:
@rwlogel cool. Want to rebase this? I'll look at this week or early next week. I want to merge #381 today or tomorrow and then release 2.1.1. Then I'd be happy to look at this thoroughly and merge it into 2.2.0 next week (which will also include support for bulk_create and queryset updates that include history)

rwlogel:
@rossmechanic rebase complete.

rossmechanic:
Thanks @rwlogel . I'll check this out early next week

simple_history/models.py
```diff
@@ -24,19 +24,28 @@
...
+
 class HistoricalRecords(object):
     thread = threading.local()
 
     def __init__(self, verbose_name=None, bases=(models.Model,),
                  user_related_name='+', table_name=None, inherit=False,
-                 excluded_fields=None,
-                 history_id_field=None,
+                 excluded_fields=None, history_id_field=None,
+                 user_model=None, get_user=default_get_user,
```
rossmechanic: Want to add `user_model` and `get_user` after `history_change_reason_field` in case anyone is just using ordered args?

simple_history/models.py
```diff
@@ -74,15 +83,11 @@ def save_without_historical_record(self, *args, **kwargs):
...
-        try:
-            hint_class = self.cls
-        except AttributeError:  # called via `register`
-            pass
-        else:
-            if hint_class is not sender:  # set in concrete
-                inherited = (self.inherit and issubclass(sender, hint_class))
-                if not inherited:
-                    return  # set in abstract
+        if self.cls is not sender:  # set in concrete
```
rossmechanic: Will this still work when you call register?

simple_history/models.py
rwlogel: Sure I can re-order them

simple_history/models.py
rwlogel: Yes I updated the register method to store this and the unit tests have been updated to verify it.

rossmechanic: What about `populate_history`? I noticed this removed adding the model to `registered_models` which is used there

rwlogel: The register function calls calls `records.finalize(model)` on line 34.  The `finalize` method calls `create_history_model` which also updates `registered_models` on line 148.

simple_history/tests/models.py
```diff
@@ -415,6 +415,71 @@ class InheritTracking4(TrackedAbstractBaseA):
...
+
+
+def get_bucket_member2(request, **kwargs):
+    try:
+        return request.user.bucket_member
+    except AttributeError:
+        return None
+
+
+class BucketDataRegister2(models.Model):
```
kseever: I see the precedent for the `..1` `..2` naming already in the file but it's not very communicative. It looks like they're differentiated by how the history user is fetched - can we make the naming reflect that?

simple_history/models.py
```diff
@@ -349,12 +356,14 @@ def get_history_user(self, instance):
...
             try:
                 if self.thread.request.user.is_authenticated:
-                    return self.thread.request.user
-                return None
+                    request = self.thread.request
             except AttributeError:
-                return None
+                pass
+
+        return self.get_user(history=self, instance=instance, request=request)
```
rossmechanic: Why include the HistoricalRecords instance in `self.get_user`? What might need to use that?

rossmechanic: Once I hear back about this `history` here, I'm happy to merge.

simple_history/tests/tests/test_admin.py
```diff
@@ -301,6 +303,21 @@ def test_deleteting_user(self):
         historical_poll = poll.history.all()[0]
         self.assertEqual(historical_poll.history_user, None)
 
+    def test_deleteting_member(self):
```
rossmechanic: typo

simple_history/models.py
rwlogel: I thought it may be useful to give the historical object to the method since it is a generic API that the user is providing.  It could be removed.

rossmechanic: Yea I don't see a reason for it to be included. Let's get rid of it for now and if someone ends up needing it, we can reevaluate. Just remove that and then I'm happy to merge.

PR #379, opened 2018-04-25
Feature/issue 378 history change reason

krisneuharth:
WIP - DO NOT MERGE!

This is my attempt at making the history_change_reason field more flexible to account for vary text lengths.

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -239,6 +239,21 @@ class HistoricalRecords(object):
                 Q(history_date__lt=self.history_date)
             ).order_by('history_date').last()
 
+        def get_history_change_reason(self):
+            cls = getattr(
+                settings,
+                'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD',
+                models.CharField
+            )
+
+            max_length = getattr(
+                settings,
+                'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_MAX_LENGTH',
+                100
+            )
+
+            return cls(max_length=max_length, null=True)
+
         if self.history_id_field:
             history_id_field = self.history_id_field
             history_id_field.primary_key = True
@@ -253,8 +268,7 @@ class HistoricalRecords(object):
         return {
             'history_id': history_id_field,
             'history_date': models.DateTimeField(),
-            'history_change_reason': models.CharField(max_length=100,
-                                                      null=True),
+            'history_change_reason': property(get_history_change_reason),
             'history_user': models.ForeignKey(
                 user_model, null=True, related_name=self.user_related_name,
                 on_delete=models.SET_NULL),
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -240,7 +240,7 @@ class HistoricalRecords(object):
             ).order_by('history_date').last()
 
         def get_history_change_reason(self):
-            cls = getattr(
+            field_cls = getattr(
                 settings,
                 'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD',
                 models.CharField
@@ -252,7 +252,7 @@ class HistoricalRecords(object):
                 100
             )
 
-            return cls(max_length=max_length, null=True)
+            return field_cls(max_length=max_length, null=True)
 
         if self.history_id_field:
             history_id_field = self.history_id_field
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -31,6 +31,7 @@ Authors
 - Jonathan Sanchez
 - Josh Fyne
 - Klaas van Schelven
+- Kris Neuharth
 - Maciej "RooTer" Urbański
 - Martin Bachwerk
 - Marty Alchin


krisneuharth:
What do you think about that approach?

The change breaks almost every test with this error:

File "/Users/kneuharth/projects/django-simple-history/simple_history/models.py", line 314, in post_save
    self.create_historical_record(instance, created and '+' or '~')
  File "/Users/kneuharth/projects/django-simple-history/simple_history/models.py", line 329, in create_historical_record
    history_change_reason=history_change_reason, **attrs)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/query.py", line 415, in create
    obj = self.model(**kwargs)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/base.py", line 495, in __init__
    raise TypeError("'%s' is an invalid keyword argument for this function" % kwarg)
TypeError: 'history_change_reason' is an invalid keyword argument for this function

Which I am still looking into but before I went too far down the path I figured I should get a conversation going.

simple_history/models.py
```diff
@@ -265,10 +252,25 @@ def get_history_change_reason(self):
         else:
             history_id_field = models.AutoField(primary_key=True)
 
+        if self.history_change_reason_field:
+            # User specific field from init
+            history_change_reason_field = self.history_change_reason_field
+            history_change_reason_field.null = True
```
rossmechanic: I don't think we have to do this. If someone wants to make their change_reason nullable with an override, I don't see a reason not to allow it.

simple_history/models.py
```diff
@@ -265,10 +252,25 @@ def get_history_change_reason(self):
...
+        elif getattr(settings, 'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD', False):
+            # User specific field from settings
+            history_change_reason_field = getattr(settings, 'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD')
+            history_change_reason_field.null = True
+            history_change_reason_field.max_length = 500
```
rossmechanic: You should probably allow the `max_length` to be a setting as well. And leave it as 100 by default so that users don't have to makemigrations with the next release if they don't need this functionality

simple_history/models.py
```diff
@@ -265,10 +252,25 @@ def get_history_change_reason(self):
...
+        else:
+            # Current default
+            history_change_reason_field = models.TextField(
```
rossmechanic: Current default is a `CharField` not a `TextField`

simple_history/registry_tests/tests.py
```diff
@@ -101,8 +101,8 @@ def test_tracked_abstract_base(self):
                 for f in TrackedWithAbstractBase.history.model._meta.fields
             ],
             [
-                'id', 'history_id', 'history_date',
-                'history_change_reason', 'history_user_id',
+                'id', 'history_id',
+                'history_change_reason', 'history_date', 'history_user_id',
```
rossmechanic: Why the ordering difference?

krisneuharth: I'm not sure yet. That was actually a question I had for you?

simple_history/models.py
```diff
@@ -30,12 +30,14 @@ class HistoricalRecords(object):
     def __init__(self, verbose_name=None, bases=(models.Model,),
                  user_related_name='+', table_name=None, inherit=False,
                  history_id_field=None,
+                 history_change_reason_field=None,
                  excluded_fields=None):
```
kseever: I don't think anyone will use all of these args positionally but if someone is, existing callers of `excluded_fields` will break.

rossmechanic: Good point. Didn't catch this in #368 but we should probably reorder it so that `excluded_fields` comes before `history_id_field` and `history_change_reason_field`

simple_history/registry_tests/tests.py
rossmechanic: I'll look into it. Not sure why but let me know if you come up with something. I'd want to know why this is happening before merging

simple_history/models.py
```diff
@@ -250,11 +252,24 @@ def get_prev_record(self):
...
             history_id_field = models.AutoField(primary_key=True)
 
+        if self.history_change_reason_field:
+            # User specific field from init
+            history_change_reason_field = self.history_change_reason_field
+        elif getattr(settings, 'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD', False):
+            # User specific field from settings
+            history_change_reason_field = getattr(settings, 'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD')
+            max_length = getattr(settings, 'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD', 100)
+            history_change_reason_field.max_length = max_length
```
rossmechanic: Actually you're right @krisneuharth . This should be nullable by default and if someone passes in a change reason field, they can make that not nullable if they'd like.

krisneuharth:
Latest round of changes with the suggested changes made.

I added a few tests but three are failing and I'm not sure why.

```Traceback (most recent call last):
  File "/Users/kneuharth/projects/django-simple-history/simple_history/tests/tests/test_models.py", line 425, in test_textfield_history_change_reason2
    TextFieldChangeReasonModel2.objects.create(greeting="what's up?")
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/query.py", line 417, in create
    obj.save(force_insert=True, using=self.db)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/base.py", line 729, in save
    force_update=force_update, update_fields=update_fields)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/base.py", line 769, in save_base
    update_fields=update_fields, raw=raw, using=using,
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/dispatch/dispatcher.py", line 178, in send
    for receiver in self._live_receivers(sender)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/dispatch/dispatcher.py", line 178, in <listcomp>
    for receiver in self._live_receivers(sender)
  File "/Users/kneuharth/projects/django-simple-history/simple_history/models.py", line 316, in post_save
    self.create_historical_record(instance, created and '+' or '~')
  File "/Users/kneuharth/projects/django-simple-history/simple_history/models.py", line 332, in create_historical_record
    history_change_reason=history_change_reason, **attrs)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/query.py", line 415, in create
    obj = self.model(**kwargs)
  File "/Users/kneuharth/projects/django-simple-history/.eggs/Django-2.0.4-py3.5.egg/django/db/models/base.py", line 495, in __init__
    raise TypeError("'%s' is an invalid keyword argument for this function" % kwarg)
TypeError: 'history_change_reason' is an invalid keyword argument for this function```

I'm not sure what is going on there because it seems like `history_change_reason` is a valid keyword arg to this:

```    def create_historical_record(self, instance, history_type):
        history_date = getattr(instance, '_history_date', now())
        history_user = self.get_history_user(instance)
        history_change_reason = getattr(instance, 'changeReason', None)

        manager = getattr(instance, self.manager_name)
        attrs = {}
        for field in self.fields_included(instance):
            attrs[field.attname] = getattr(instance, field.attname)
        manager.create(history_date=history_date, history_type=history_type,
                       history_user=history_user,
                       history_change_reason=history_change_reason, **attrs)```

I'll start working on the documentation changes today, but in the mean time does anyone have any idea what is going on there?

**Comment by rossmechanic** [comment]
Which tests is it failing? My guess is it's only failing the tests in which you set `TextField` in the settings. You're better off making the setting `'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD_TEXT_FIELD'` a boolean, and instantiating the field as a `TextField` if that boolean is true. As it stands now, you set `TextField` in the settings without ever instantiating the field. So the historical model never gets a `history_change_reason` and the manager doesn't expect that field.

File: simple_history/tests/models.py [hunk 7]
```diff
@@ -432,3 +432,28 @@ class UUIDDefaultModel(models.Model):
 
 # Clear the SIMPLE_HISTORY_HISTORY_ID_USE_UUID
 delattr(settings, 'SIMPLE_HISTORY_HISTORY_ID_USE_UUID')
+
+
+# Set the SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD and MAX_LENGTH
+setattr(settings, 'SIMPLE_HISTORY_HISTORY_CHANGE_REASON_FIELD', models.TextField)
```
rossmechanic: Here's the line. You don't instantiate the `TextField` here. And you don't instantiate it when you call the setting either. That's why the field isn't expected.

krisneuharth: Oh, makes sense. Thanks!

simple_history/registry_tests/tests.py
krisneuharth: I think this is happening in `create_history_model` where it stores the history fields in the attrs dict. I think by moving the positional args in the __ini__ method it changed the natural ordering of attrs. I don't think this is a problem. What do you think?

simple_history/models.py
rossmechanic: @krisneuharth see @kseever's comment here about positional args. Should probably add this after the other positional args.

rossmechanic:
@krisneuharth tests are still failing. Want to update this?

krisneuharth:
@rossmechanic I made the requested changes, tests added, all tests pass, and usage documentation has been updated. Thanks for reviewing!

krisneuharth:
Just checking in on this PR. @rossmechanic do I need to make any more changes? Thanks!

rossmechanic:
@krisneuharth sorry hadn't seen the changes. Busy week but I will try to get to this this week

PR #385, opened 2018-05-07
Evaluate method returned from getattribute, if it is callable

blawson:
Currently, if you have a method defined in history_list_display, None is returned as the method isn't evaluated from getattribute.

This should allow folks to define methods in their attribute lists for history.


rossmechanic:
@blawson Thanks for the change. 
Does this account for callables that are on the admin? How do we account for ForeignKeys? The historical object won't have the same functionality as the current object. Worried about the case where a callable traverses a foreign key relationship, and that relationship doesn't exist on the historical object. I guess that will just error out and the user can avoid using those callables. However, still would like the functionality to work for callables on the admin as well.

PR #388, opened 2018-05-11
Fix middleware cleanup failure

### LINKED ISSUE #387: Middleware cleanup failure
tl;dnr I think this block of code should be in a finally clause
https://github.com/treyhunner/django-simple-history/blob/149911efe7bc96237186625ed5914aa3b44815dc/simple_history/middleware.py#L26

I'm getting an integrity error in the setup for a test. Basically whats happening is the preceding test exceptions out of a view. Subsequently this middleware cleanup never runs and simple history hangs onto the user reference. Meanwhile test cleanup removes the user from the database. Then during setup for the next test we are creating objects outside a view an simple history tries to use the stored user which causes an integrity error in the database.

adamchainz:
Fixes #387 using the `patchy.patch` provided on the issue.

The test isn't perfect, it passes before applying the fix, though I can't work out why.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Unreleased
 ----------
 - Add ability to specify custom history_id field (gh-368)
 - Add HistoricalRecord instance properties `prev_record` and `next_record` (gh-365)
+- Fix exception handling in ``HistoryRequestMiddleware``
 
 2.0 (2018-04-05)
 ----------------
--- a/simple_history/middleware.py
+++ b/simple_history/middleware.py
@@ -13,17 +13,9 @@ class HistoryRequestMiddleware:
         self.get_response = get_response
 
     def __call__(self, request):
-        # Code to be executed for each request before
-        # the view (and later middleware) are called.
-
         HistoricalRecords.thread.request = request
-
-        response = self.get_response(request)
-
-        # Code to be executed for each request/response after
-        # the view is called.
-
-        if hasattr(HistoricalRecords.thread, 'request'):
-            del HistoricalRecords.thread.request
-
-        return response
+        try:
+            return self.get_response(request)
+        finally:
+            if hasattr(HistoricalRecords.thread, 'request'):
+                del HistoricalRecords.thread.request
--- a/simple_history/tests/tests/test_middleware.py
+++ b/simple_history/tests/tests/test_middleware.py
@@ -4,6 +4,7 @@ from django.conf import settings
 from django.test import TestCase, override_settings
 from django.urls import reverse
 
+from simple_history.models import HistoricalRecords
 from simple_history.tests.custom_user.models import CustomUser
 from simple_history.tests.models import Poll
 
@@ -13,6 +14,7 @@ overridden_settings = {
 }
 
 
+@override_settings(**overridden_settings)
 class MiddlewareTest(TestCase):
     def setUp(self):
         self.user = CustomUser.objects.create_superuser(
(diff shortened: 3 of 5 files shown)


adamchainz:
cc @tolomea

rossmechanic:
Hey @adamchainz thanks for the fix. I don't love the idea of merging this without a test that fails before the change. The reason an exception doesn't hit the middleware (and why the middleware still deletes the request on the thread even with the exception in the test) is here https://github.com/django/django/blob/master/django/core/handlers/exception.py#L18. `convert_exception_to_response` is run on all Django middleware. Not sure why the same function wouldn't be running in #387. @tolomea , any thoughts?

tolomea:
It's Django 1.11 and the exception I hit this on was an AssertionError
(test code).

tolomea:
During test it goes through this line instead:
https://github.com/django/django/blob/a5a2ceeb453b86caaf01687342984f372428beb8/django/core/handlers/exception.py#L121

And so the comment in the source that "This decorator is automatically
applied to all middleware to ensure that no middleware leaks an exception
and that the next middleware in the stack can rely on getting a response
instead of an exception." does not hold during testing.

rossmechanic:
@tolomea ok that is fine with me. But I'd still want a test that fails without the change and passes with the change before acceptance.

tolomea:
I did some more digging into this.
It turns out that that DEBUG_PROPAGATE_EXCEPTIONS flag is not set True by the Django test runner or pytest-django and it's not implied by having DEBUG set True.
Where it's getting set True for me is in django-webtest.
You can make the test fail (on the existing code) by using override settings to turn that flag on.

I'm not convinced that fixing this in your middleware is the right answer. On the one hand this is getting kinda specific to certain test environments. On the other hand when it fails it's a real nuisance to track down.
Thoughts?

tolomea:
It occurs to me that it'd be kinda nice if there was a start or end of test signal, then something like this could attach to it and make sure the environment is sane.

rossmechanic:
@tolomea Thanks for digging into this! Yes, I agree. I think it's too specific a use case to include in the DSH middleware. What do you think about listing it under the Common Issues section in the docs? That way other users who run into this issue know why it happens. For your specific tests, you could just have a `tearDown` method in your tests that removes the request from the thread. What do you think?

tolomea:
I already have a global test cleanup spot that I use for stuff like flushing the task queue and caches, it's easy enough for me to add it there.

Documenting it definitely a good idea. It manifests as a IntegrityError on a User FK when adding a row to a history table.

django.db.utils.IntegrityError: (1452, 'Cannot add or update a child row: a foreign key constraint fails (`test_env`.`core_historicaladdress`, CONSTRAINT `core_historicaladdress_history_user_id_0f2bed02_fk_user_user_id` FOREIGN KEY (`history_user_id`) REFERENCES `user_user` (`id`))')

Basically you need a webtest call, followed by a database rollback, followed by an ORM operation that creates a row in history but is not inside a request.

Given the nature of the various bits involved this seems likely to usually be in a similar situation to mine where one test finished with a webtest call and the next test blows up while creating fixtures in test setup.
And subsequently it's prone to come and go as the test order changes.

rossmechanic:
So then I would just close this PR and put up a new on that mentions the testing issues that come when you use the DSH middleware and use webtest for testing. I'm happy to add that to the documentation, or you're welcome to add it there. Let me know if you're going to add the doc PR. Thanks!

tolomea:
If you don't mind I think it will be quicker and fit with the existing
documentation better if you do it.

rossmechanic:
All good. Thanks for your help.

adamchainz:
👍 Well diagnosed

PR #390, opened 2018-05-15
Set attribute on history list entries, from evaluated admin methods

blawson:
This change will allow admin pages that extend SimpleHistoryAdmin to provide admin methods that will be evaluated on each object in the change list.

I.e., you can update history_list_display with a method, defined in the admin, that will be evaluated and set on each object.


blawson:
Adding tests.

rossmechanic:
Failing flake8 style @blawson

simple_history/admin.py
```diff
@@ -61,6 +61,13 @@ def history_view(self, request, object_id, extra_context=None):
         if not self.has_change_permission(request, obj):
             raise PermissionDenied
 
+        # Set attribute on each action_list entry from admin methods
+        for x in history_list_display:
```
rossmechanic: change from `x` to a more descriptive var name

PR #392, opened 2018-05-15
Added documentation for failures resulting from django-webtest

### LINKED ISSUE #387: Middleware cleanup failure

rossmechanic:
@tolomea let me know if there's anything you want to add to this. 
Fixes #387

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -1,6 +1,9 @@
 Common Issues
 =============
 
+Tracking Custom Users
+---------------------
+
 -   ``fields.E300``::
 
         ERRORS:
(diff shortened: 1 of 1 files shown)


tolomea:
I think that about covers it.

PR #408, opened 2018-06-14
Fix out of memory exception in populate_history management command

### LINKED ISSUE #239: populate_history uses an incorrect Manager
The populate_history command assumes (in `bulk_history_create`) that the Manager for the Model is `objects`, while it may be anything. This causes trouble with models that redefine the meaning of `objects` or remove it altogether. Instead, the command should use `Model._default_manager` as per https://docs.djangoproject.com/en/dev/topics/db/managers/#default-managers.
The default manager is used inter alia by the Django Admin, so it makes sense to use it here too.

rossmechanic:
Thanks to @markddavidoff for #355 which this PR closes. This also closes #239 by using the `_default_manager` rather than `objects` when querying models. 
## Description

This fix makes sure that `populate_history` reads from model tables in batches in order to prevent out-of-memory exceptions caused by reading too many rows into memory. This also refactors `bulk_history_create` so that it is now on the `history` manager. 

## Motivation and Context

The `populate_history` management command used to work by finding each model that has a historical table, checking to see if the historical table is empty, and if so, reading every record from the model's table into memory, to then bulk_create the history. People were running into issues where they'd try to run `populate_history` on a table with millions of rows, resulting in all of those rows being read into memory simultaneously, which led to out-of-memory exceptions. 

## How Has This Been Tested?

Added tests for the `bulk_history_create` function on the `history` manager (in `simple_history/tests/tests/test_manager.py`) and added tests in `simple_history/tests/tests/test_commands.py`. 

## Types of changes

## Checklist:


simple_history/management/commands/populate_history.py
```diff
@@ -54,14 +57,16 @@ def handle(self, *args, **options):
             for model in models.registered_models.values():
                 try:    # avoid issues with mutli-table inheritance
                     history_model = utils.get_history_model_for_model(model)
-                except utils.NotHistorical:
+                except NotHistoricalModelError:
                     continue
                 to_process.add((model, history_model))
             if not to_process:
-                self.stdout.write(self.NO_REGISTERED_MODELS)
+                if self.verbosity >= 1:
```
kseever: Can we get constants for the verbosity levels? This is magic AF

rossmechanic: It's actually Django AF. https://github.com/django/django/blob/master/django/core/management/commands/makemigrations.py#L167

simple_history/management/commands/populate_history.py
```diff
@@ -54,14 +57,16 @@ def handle(self, *args, **options):
...
+                    self.stdout.write(self.NO_REGISTERED_MODELS)
```
kseever: Is `stdout` the usual convention for Django library logging?

rossmechanic: Yep

simple_history/manager.py
```diff
@@ -88,3 +89,19 @@ def _as_of_set(self, date):
...
+    def bulk_history_create(self, objs, batch_size=None):
+        """Bulk create the history for the objects specified by objs"""
+
+        historical_instances = [
+            self.model(
+                history_date=getattr(instance, '_history_date', now()),
+                history_user=getattr(instance, '_history_user', None),
+                **{
+                    field.attname: getattr(instance, field.attname)
+                    for field in instance._meta.fields
```
kseever: Is this where you need to account for `excluded_fields`?

rossmechanic: Yes. Going to get this merged and then ask @uadnan to rebase since we're still thinking about whether the solution there is best way to do things.

simple_history/tests/tests/test_manager.py
```diff
@@ -72,17 +76,71 @@ class AsOfAdditionalTestCase(TestCase):
...
+        self.data = [
+            Poll(id=1, question='Question 1', pub_date=datetime.now()),
+            Poll(id=2, question='Question 2', pub_date=datetime.now()),
+            Poll(id=3, question='Question 3', pub_date=datetime.now()),
+            Poll(id=4, question='Question 4', pub_date=datetime.now()),
+        ]
+
+    def test_simple_bulk_history_create(self):
+        created = Poll.history.bulk_history_create(self.data)
+        self.assertEqual(len(created), 4)
```
kseever: `self.assertEqual(len(created), len(self.data))` would be less brittle in the event that we ever change the number of items in there.

simple_history/tests/tests/test_manager.py
```diff
@@ -72,17 +76,71 @@ class AsOfAdditionalTestCase(TestCase):
...
+        self.assertQuerysetEqual(Poll.history.order_by('question'), [
+            'Question 1', 'Question 2', 'Question 3', 'Question 4'
+        ], attrgetter('question'))
+
+        created = Poll.history.bulk_create([])
```
kseever: This feels odd here. What exactly are you testing? Kind of feels like it should be its own testcase.

simple_history/tests/tests/test_manager.py
rossmechanic: Was following how django tests their bulk_create https://github.com/django/django/blob/master/tests/bulk_create/tests.py#L28

And also prefer to have constants in the expected value

simple_history/tests/tests/test_manager.py
rossmechanic: Fair. I'll move it

simple_history/tests/tests/test_manager.py
```diff
@@ -72,17 +76,71 @@ class AsOfAdditionalTestCase(TestCase):
...
+            Poll(question='Question 2', pub_date=datetime.now()),
+            Poll(question='Question 3', pub_date=datetime.now()),
+            Poll(question='Question 4', pub_date=datetime.now()),
+        ]
+
+        with self.assertRaises(IntegrityError):
+            Poll.history.bulk_history_create(self.data)
+
+    def test_set_custom_history_date_on_first_obj(self):
+        self.data[0]._history_date = datetime(2000, 1, 1)
```
kseever: Y2K?

rossmechanic: First date that came to mind 🙃

markddavidoff:
@rossmechanic thanks again

PR #417, opened 2018-07-30
Document some register attributes

dgilge:
Addressing several attributes of `register` for use with third-party apps:
1. `app`
2. `excluded_fields`
3. Reference for other attributes

I needed this and searched for a clean solution (almost opened an issue) and was happy when I saw that it is implemented already. People should know that all these features exist! :)

I implemented it in my project and it works. I only changed the documetation. Therefore I assume that I don't have to write a test.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -64,6 +64,23 @@ third-party apps you don't have control over.  Here's an example of using
 
     register(User)
 
+If you want to separate the migrations of the historical model from those of
+the third-party model itself you can pass a module as ``app`` attribute to
+``register``. For instance, if the migrations shall live in the migrations
+folder of the package you register the model in, you could do:
+
+.. code-block:: python
+
+    register(User, app=__package__)
+
+You can pass attributes of the ``HistoricalRecords`` model directly to ``register``:
+
+.. code-block:: python
+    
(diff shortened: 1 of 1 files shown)


docs/advanced.rst
```diff
@@ -428,4 +445,4 @@ This may be useful when you want to construct timelines and need to get only the
     new_record, old_record = p.history.all()
     delta = new_record.diff_against(old_record)
     for change in delta.changes:
-        print("{} changed from {} to {}".format(change.field, change.old, change.new))
\ No newline at end of file
+        print("{} changed from {} to {}".format(change.field, change.old, change.new))
```
rossmechanic: I don't see a diff here? Spacing?

rossmechanic:
Thanks for updating the docs! Just one comment – looks like there was no change (spacing?). Might want to just get rid of that and then I'll merge.

docs/advanced.rst
dgilge: GitHub adds a blank line at the end. I'm going to fix it.

dgilge: No, it doesn't. I'm afraid I don't know how to fix this.

rossmechanic: No worries. Merging. Thanks!

PR #422, opened 2018-08-15
Add IP Address to history record

mscansian:
Adds the request's IP Address to the history record and updates the admin template to show this new information. I'm using the header `HTTP_X_FORWARDED_FOR` as the IP's source if available, otherwise, if falls back to `REMOTE_ADDR`.

## Description
* Adds a new field: `history_remote_addr`
* Changes `create_historical_record` to pass the IP Address from the request to the history record.
* Updates the admin template to show the IP address
* Adds tests to the new functionality and update a few tests that broke
* Updates the pt_BR translation to include the new functionality

## Related Issue
https://github.com/treyhunner/django-simple-history/issues/421

## Motivation and Context
As described in the issue above, we use this project as an audit tool. We need to have not only the user that made a changed, but also which IP address was he using at the time of the change. Also, I think that this makes a nice addition for a lot of use cases :)

## How Has This Been Tested?
I'm testing the middleware providing a reverse proxy header `HTTP_X_FORWARDED_FOR` and testing if the record includes the IP address. I'm also doing the same test only with `REMOTE_ADDR`

screenshot from 2018-08-15 17-35-51

## Types of changes

- [X] Requires migration (all users need to run makemigrations/migrate)

## Checklist:

- [X] My code follows the code style of this project.


simple_history/models.py
```diff
@@ -31,6 +31,13 @@ def default_get_user(request, **kwargs):
         return None
 
 
+def get_remote_addr(request):
+    x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
+    if x_forwarded_for:
+        return x_forwarded_for.split(',')[0]
```
ThePumpingLemma: Should we also log the remainder of the chain and the IP of the host that sent the request to Django given that it is easier to spoof IPs at the beginning of the chain?

mscansian: Haven't thought about that, it might be a good idea! Especially in my use case. I'm just not sure how to do it without over-complicating things for the average user. After all, 99% of the time all you want is the first address in the chain.

mscansian:
@kseever, as a simpler solution, we could use [django-ipware](https://github.com/un33k/django-ipware) to avoid dealing with HTTP altogether.

The only drawback I can see is that it adds a dependency. Though it's likely that other libs in most Django apps also depends on this one (notably [django-axes](https://github.com/jazzband/django-axes/blob/master/setup.py#L29)).

vdboor:
@mscansian I've stated this in multiple projects, I consider django-ipware to be a security risk. It allows various headers by default making the app susceptible to IP-spoofing. A secure solution is using wsgiunproxy / aws-wsgi-unproxy that rewrites the `REMOTE_ADDR` header to point to the proper IP address - but only when it came from a trusted proxy. This also fixes any other issues with Sentry or rate limiting checks or other packages that all assume `REMOTE_ADDR` is the right thing to do.

mscansian:
Found this [rather long discussion](https://github.com/jazzband/django-axes/issues/286) about the same issue on django-axes. In the end, they decided to just use `REMOTE_ADDR`. Though, this is only the default setting ([as described in the docs](https://django-axes.readthedocs.io/en/latest/configuration.html#reverse-proxy-configuration)). `django-ipware` is still used in the project to allow people that ~know what they are doing~ want to use another configuration to do so.

We could probably do something similar or be even more conservative. I just want to avoid yet another 3 months discussion on this matter.

rossmechanic:
@mscansian just to update on this. We're working on an interface to easily add columns to the history table, and should have something out in the next two weeks. Once that is done, I'm thinking that we have IP address logging as a pluggable feature that you can turn on in settings.

rossmechanic:
This should be covered by #471 . Now it should be super easy for you to add fields to the history table and add the IP address. Going to close this PR

Issue #439, opened 2018-09-28
Allow historical data to be deleted when the master record is deleted


rwlogel:
**Problem Statement**
We have a number of tables using `HistoricalRecords`,  when attempting to delete a user in the admin it fails (see ticket #391).  In our circumstance if the record is being deleted we don't actually want to keep any historical records.

**Describe the solution you'd like**
I'd like to be able to specify that the historical records would be deleted when the master record is deleted.

PR #440, opened 2018-09-28
Add feature to allow historical records to be deleted when master is deleted

rwlogel:
## Description
Allow user to specify that they don't want to keep history for deleted models.

## Related Issue
#439 

## Motivation and Context
If users don't want to keep history for deleted data this allows them to do so.

## How Has This Been Tested?
The unit tests have been updated to account for the new functionality.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Changes
 =======
 
+Unreleased
+----------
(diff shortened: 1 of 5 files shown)


ThePumpingLemma:
@rossmechanic I think the constraint would be more efficient, but I think we can do that as a follow-on PR later.

PR #461, opened 2018-10-19
Fix custom user bug and update to 2.5.1

pm13:
Another variant of https://github.com/treyhunner/django-simple-history/pull/460

## Description
See https://github.com/treyhunner/django-simple-history/pull/460 - but it is not full revert of https://github.com/treyhunner/django-simple-history/commit/372632779592b026756cbfd83fc8712afb7a6978

## Related Issue
Fixes https://github.com/treyhunner/django-simple-history/issues/459

## How Has This Been Tested?

It solves problem described in https://github.com/treyhunner/django-simple-history/issues/459 - I try this solution before creating of issue.

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,12 @@
 Changes
 =======
 
-Unreleased
------------
+2.5.1 (2018-10-19)
+------------------
 - Add `'+'` as the `history_type` for each instance in `bulk_history_create` (gh-449)
 - Add support for  `history_change_reason` for each instance in `bulk_history_create` (gh-449)
-- Add `history_change_reason` in the history list view under the  `Change reason` display name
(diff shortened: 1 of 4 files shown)


rossmechanic:
I tried this locally and it failed so I did a full revert. Are you not seeing the same failure?

pm13:
You are right - my use case is more complex (more user models with inheritance). And I didn't know that this code should work:

```python
from django.contrib.auth.models import AbstractUser
from simple_history.models import HistoricalRecord

class User(AbstractUser):
     history = HistoricalRecords()
```
(It works only with the full revert.)

This code works with both patches:

```python
from django.contrib.auth.models import AbstractUser
from simple_history import register

class User(AbstractUser):
    pass

register(User)
```
(I think it was necessary with old version of Django?)

PR #462, opened 2018-10-20
Ensure custom arguments for fields are included in historical models' fields

### LINKED ISSUE #431: django-simple-history does not respect 'deconstruct' kwargs for custom foreignkey fields
**Describe the bug**
After subclassing ForeignKey to override get_attname, and defining a deconstruct method to carry the additional parameter, makemigrations creates the correct migrations but django-simple-history ignores the extra parameters.  The resulting models are unusable.  A custom migration fixes the issue but will be replaced on the next makemigrations operation. 

The custom foreign key field is motivated and inspired by:
https://stackoverflow.com/questions/21524830/modifying-djangos-default-id-suffix-for-related-objects/24427909#24427909

**To Reproduce**

Deconstruct, as defined in Django docs [here](https://docs.djangoproject.com/en/2.1/howto/custom-model-fields/#field-deconstruction), is used to carry the attr_name argument into the migrations.  Here is the class: 
```
class CustomAttrNameForeignKey(models.ForeignKey):
    def __init__(self, *args, **kwargs):
        self.attr_name = kwargs.pop('attr_name', None)
        super().__init__(*args, **kwargs)

    def get_attname(self):
        return self.attr_name or super().get_attname()

    def deconstruct(self):
        name, path, args, kwargs = super().deconstruct()
        if self.attr_name:
            kwargs['attr_name'] = self.attr_name
        return name, path, args, kwargs
```

**Expected behavior**
When running make migration, Django creates 2 migrations as expected.  One for the model and one for the historical model.  However, the migration associated with the historical model does not have the needed 'attr_name' argument.  This makes all operations (save, get_or_create, etc.) on model instances fail.

Here is an example of the generated migration code:
```
        migrations.AlterField(
            model_name='mymodel',
            name='status_fk', 
            field=CustomAttrNameForeignKey(attr_name='status', db_column='status', to='MyStatusModel'),
        ),
        migrations.AlterField(
            model_name='historicalmymodel',
            name='status_fk',
            field=CustomAttrNameForeignKey(blank=True, db_column='status', db_constraint=False, null=True, on_delete=django.db.models.deletion.DO_NOTHING, related_name='+', to='MyStatusModel'),
        ),
```

 - OS: Ubuntu 16.04
 - Django Simple History Version: 2.1.1
 - Django Version: 1.11.6
 - Database Version: PostgreSQL 9.6.6

tbeadle:
## Description
Previously, if a custom field class was used for a foreign key and that
class accepted custom kwargs, those kwargs would not be included in the
creation of the historical model's field and would break the migration.
To address this, we'll use the deconstruct() method of the field to get
the arguments that should be passed to it and override the ones we need
to for the historical model's field.

## Related Issue
Fixes #431 

## Motivation and Context
See description

## How Has This Been Tested?
Added unit tests

## Types of changes

## Checklist:
- [X] My code follows the code style of this project.


simple_history/models.py
```diff
@@ -178,29 +178,18 @@ def copy_fields(self, model):
                 field.__class__ = models.IntegerField
             if isinstance(field, models.ForeignKey):
                 old_field = field
-                field_arguments = {'db_constraint': False}
+                old_field.swappable = False
+                name, path, args, field_arguments = old_field.deconstruct()
```
rossmechanic: Make `name` and `path` underscores since they aren't used?

simple_history/models.py
```diff
@@ -178,29 +178,18 @@ def copy_fields(self, model):
                 field.__class__ = models.IntegerField
             if isinstance(field, models.ForeignKey):
                 old_field = field
-                field_arguments = {'db_constraint': False}
+                old_field.swappable = False
```
rossmechanic: Let's change this back to its original setting after calling deconstruct

rossmechanic: @tbeadle let's change the swappable setting back after doing this deconstruct? And I'd love to figure out why those fields need to be sorted in tests – haven't been able to find why the order changes but want to know before merging. I'll look into it some more

PR #457, opened 2018-10-19
Issue 341 - Warning if HistoricalRecords(inherit=False) in an abstract class

shlywa:
## Description
Working on this at a sprint at DjangoCon 2018.

I am issuing a Python Error here with warnings.warn. The other option I see is to try to issue it with Django's System Check Framework, but I'm not 100% sure how to do that, since it would probably require extending the check method, but then at that point I'm not sure how I would go about getting the cls object.

## Related Issue
https://github.com/treyhunner/django-simple-history/issues/341

## Motivation and Context
If HistoricalRecords is added to an abstract model and inherit=False, the models that are subclassed from this abstract model will not have their histories saved.

## How Has This Been Tested?
I ran a sample project on Django 2.1 that tried to create an abstract model with HistoricalRecords(inherit=False) and the warning was successfully raised. I also tried to subclass that abstract model and the warning was raised for that subclassed model as well.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

I'm not 100% sure how to run further tests, nor am I sure if raising a Python-based exception is the in code style of this project. Please let me know and I'll change things accordingly!

PR #472, opened 2018-10-31
GH-341: Warning if HistoricalRecords(inherit=False) in an abstract class

### LINKED ISSUE #341: Warn when using HistoricalRecords on an abstract class without inherit=True
Adding a `HistoricalRecords` to an abstract model without `inherit=True` unlike most other Field types doesn't do anything. It should therefore trigger a warning with a hint to `inherit=True`.

```python
class CommonFields(models.Model):
    history = HistoricalRecords()

    class Meta:
        abstract = True

class Foo(CommonFields):
    pass
```

rossmechanic:
ADAPTED FROM #457 . Thanks @shlywa !
## Description
Working on this at a sprint at DjangoCon 2018.

I am issuing a Python Error here with warnings.warn. The other option I see is to try to issue it with Django's System Check Framework, but I'm not 100% sure how to do that, since it would probably require extending the check method, but then at that point I'm not sure how I would go about getting the cls object.

## Related Issue
Closes #341 

## Motivation and Context
If HistoricalRecords is added to an abstract model and inherit=False, the models that are subclassed from this abstract model will not have their histories saved.

## How Has This Been Tested?
I ran a sample project on Django 2.1 that tried to create an abstract model with HistoricalRecords(inherit=False) and the warning was successfully raised. I also tried to subclass that abstract model and the warning was raised for that subclassed model as well.

## Types of changes

## Checklist:
- [ ] I have added tests to cover my changes. - Tested locally. 

I'm not 100% sure how to run further tests, nor am I sure if raising a Python-based exception is the in code style of this project. Please let me know and I'll change things accordingly!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Unreleased
 ----------
 - Add `custom_model_name` parameter to the constructor of `HistoricalRecords` (gh-451)
 - Fix header on history pages when custom site_header is used (gh-448)
+- Raise warning if HistoricalRecords(inherit=False) is in an abstract model (gh-341)
 
 2.5.1 (2018-10-19)
 ------------------
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -73,8 +73,8 @@ class HistoricalRecords(object):
         self.add_extra_methods(cls)
 
         if cls._meta.abstract and not self.inherit:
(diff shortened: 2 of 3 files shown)


rossmechanic:
Going to find a way to test this so coverage isn't hurt @kseever

PR #476, opened 2018-11-02
Allow user with view permissions to view the history

dgerritsen:
## Description
Currently, a user with view permissions can see the model instance itself, but not the history.
With this change, users with view, but not change permission, can see the history, but not revert etc.

## Related Issue
https://github.com/treyhunner/django-simple-history/issues/475

## Motivation and Context
With this change, users with view, but not change permission, can see the history, but not revert etc.

## How Has This Been Tested?
Has been tested in a project i'm currently developing, and in which I noticed this.

## Types of changes

## Checklist:

Working on the tests.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -18,6 +18,7 @@ Authors
 - Daniel Levy
 - Daniel Roschka
 - David Hite
+- Dion Gerritsen
 - Eduardo Cuducos
 - Florian Eßer
 - George Vilches
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 - Fix header on history pages when custom site_header is used (gh-448)
 - Modify `pre_create_historircal_record` to pass `history_instance` for ease of customization (gh-421)
 - Raise warning if HistoricalRecords(inherit=False) is in an abstract model (gh-341)
+- Let users with view permissions see the model's history (gh-475)
 
 2.5.1 (2018-10-19)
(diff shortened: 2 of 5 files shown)


rossmechanic:
@dgerritsen Thanks for this! A few comments:
1. view permissions are only available as of django 2.1, so we need to make this backwards compatible. We support back to django 1.11. 
2. Definitely need automated tests for this

rossmechanic:
@dgerritsen are you still working on this?

dgerritsen:
Sorry, I've been a bit busy with other stuff lately. Will try to work on this the next few days.

rossmechanic:
@dgerritsen any update?

dgerritsen:
I have no clue how to make this backwards compatible, and I can't figure out how to get those tests working. Anyone willing to think with me?

rossmechanic:
@dgerritsen happy to help explain. Most of the tests are failing because `has_view_permission` is a function that only exists in Django 2.1, but we're testing on Django versions 1.11, 2.0, and 2.1. Thus, the code has to do different things depending on the version. For example, you may want something like:
```
import django
if django.VERSION >= (2, 1) then:
    return self.has_view_permission()
else:
    return self.has_change_permission()
```

However, this PR is a little more complicated, as it's not entirely clear that we want view-only users to be able to view the history. It feels like a separate permission to me. I don't want to assume that users with the view permission have the ability to view history. Going to close this PR and handle it myself because there are some weird edge cases here

PR #477, opened 2018-11-07
Most recent with excluded fields

Alig1493:
## Description
Added changes in manager to compensate for excluded fields while feetching most recent. Updated test_model_with_excluded_fields test to verify.

### Steps to reproduce: 
* Create a model with a historical records field
* Add an excluded model field in historical records
* Query for most_recent of the model's history field
https://github.com/treyhunner/django-simple-history/issues/333

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -34,6 +34,11 @@ class HistoryManager(models.Manager):
             key_name = self.instance._meta.pk.name
         return self.get_super_queryset().filter(**{key_name: self.instance.pk})
 
+    def get_excluded_fields(self):
+        if isinstance(self.instance._meta.model.history, self.__class__):
+            return self.instance._meta.model.history.model._history_excluded_fields
+        return []
+
     def most_recent(self):
         """
         Returns the most recent copy of the instance available in the history.
@@ -45,12 +50,7 @@ class HistoryManager(models.Manager):
                 )
             )
         tmp = []
-        excluded_fields = []
-
-        if isinstance(self.instance._meta.model.history, self.__class__):
-            excluded_fields = (
-                self.instance._meta.model.history.model._history_excluded_fields
-            )
+        excluded_fields = self.get_excluded_fields()
 
         for field in self.instance._meta.fields:
             if isinstance(field, models.ForeignKey):
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -51,11 +51,12 @@ class HistoryManager(models.Manager):
         excluded_fields = self.get_excluded_fields()
 
         for field in self.instance._meta.fields:
-            if field.name not in excluded_fields:
-                if isinstance(field, models.ForeignKey):
-                    tmp.append(field.name + "_id")
-                else:
-                    tmp.append(field.name)
+            if field.name in excluded_fields:
+                continue
+            if isinstance(field, models.ForeignKey):
+                tmp.append(field.name + "_id")
+            else:
+                tmp.append(field.name)
         fields = tuple(tmp)
         try:
             values = self.get_queryset().values_list(*fields)[0]
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -4,6 +4,7 @@ from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.timezone import now
 
+
 class HistoryDescriptor(object):
     def __init__(self, model):
         self.model = model
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -35,9 +35,7 @@ class HistoryManager(models.Manager):
         return self.get_super_queryset().filter(**{key_name: self.instance.pk})
 
     def get_excluded_fields(self):
-        if isinstance(self.instance._meta.model.history, self.__class__):
-            return self.instance._meta.model.history.model._history_excluded_fields
-        return []
+        return self.model._history_excluded_fields
 
     def most_recent(self):
         """
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 
 from django.db import models
-from django.utils.encoding import python_2_unicode_compatible
 from django.utils.timezone import now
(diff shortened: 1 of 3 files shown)


simple_history/models.py
```diff
@@ -65,6 +65,9 @@ def __init__(self, verbose_name=None, bases=(models.Model,),
         except TypeError:
             raise TypeError("The `bases` option must be a list or a tuple.")
 
+    def _get_excluded_fields(self):
```
rossmechanic: Is this being used? Can't see it being called anywhere else

simple_history/tests/tests/test_models.py
```diff
@@ -398,8 +398,10 @@ def test_model_with_excluded_fields(self):
         p.save()
         history = PollWithExcludeFields.history.all()[0]
         all_fields_names = [f.name for f in history._meta.fields]
+        most_recent = p.history.most_recent()
         self.assertIn('question', all_fields_names)
         self.assertNotIn('pub_date', all_fields_names)
+        self.assertEqual(most_recent.__class__, PollWithExcludeFields)
```
kseever: Could we also get a test case that ensures that the excluded fields are indeed excluded?

simple_history/models.py
Alig1493: It's not being used anywhere. My mistake since I hadn't spot the meta attribute containing the excluded fields in the first place.

simple_history/tests/tests/test_models.py
Alig1493: @kseever the test here is already checking whether the field is present in the fields of the history model so I think it's already being covered. I have added another statement checking if the field is present in the excluded fields of the history model. It might be that I may have failed to understand your requirement so please feel free to elaborate. =)

kseever: Ah, I just noticed that the assertion I was looking for already exists on line 403 - confirming that `pub_date` isn't in the list of returned fields. Nevermind me!

rossmechanic:
@Alig1493 not sure why this isn't building. Want to try fixing conflicts and pushing?

Alig1493:
I'll have the conflicts resolved. SOrry for the delayed response

rossmechanic:
Want to run `make format` and fix the build? @Alig1493

Alig1493:
@rossmechanic sure I'll do it right away. sorry for the late reply.

Alig1493:
@rossmechanic any advice regarding solving cyclomatic complexity `def most_recent(self)`?

rossmechanic:
@Alig1493 refactoring your new code into a `get_excluded_fields` function should do the trick

Alig1493:
@rossmechanic Is my latest change what you had in mind? Maybe I missed something?

Alig1493:
@rossmechanic Is my latest change what you had in mind? Maybe I missed something?

rossmechanic:
@Alig1493 yes! just rebase and the codeclimate issue should go away. However, I would like to see the code coverage increase

Alig1493:
I'll have it rebased soon. Sorry was on leave.

Alig1493:
@rossmechanic I have rebased and pushed but it seems the initial problem still persists. Anything I might have missed?

simple_history/manager.py
```diff
@@ -34,6 +34,11 @@ def get_queryset(self):
             key_name = self.instance._meta.pk.name
         return self.get_super_queryset().filter(**{key_name: self.instance.pk})
 
+    def get_excluded_fields(self):
+        if isinstance(self.instance._meta.model.history, self.__class__):
```
rossmechanic: Don't want to hardcode `history` here. Instead, use `utils.get_history_manager_for_model`

rossmechanic: What is this conditional checking for?

simple_history/manager.py
```diff
@@ -34,6 +34,11 @@ def get_queryset(self):
...
+            return self.instance._meta.model.history.model._history_excluded_fields
```
rossmechanic: could do self.model._history_excluded_fields

rossmechanic:
@Alig1493 don't worry about the codeclimate error, but I left a few other comments. Also want to get the codecov/patch up to 100%

simple_history/manager.py
Alig1493: I was using this conditional check because I thought all the instances might not have a model history class.

Alig1493:
@rossmechanic the codecov patch is upto a 100% just like you mentioned.

simple_history/manager.py
rossmechanic: They should all have a model history class. However, they might not have `_history_excluded_fields` so you should do:
`excluded_fields = getattr(self.model, '_history_excluded_fields', [])`

rossmechanic:
@Alig1493 Great! Just left one more comment

simple_history/manager.py
Alig1493: oh shoot you're right. Now that I'm looking at the print statements for debugging it's painfully obvious. Damn!

Alig1493:
@rossmechanic I've made your requested changes.

simple_history/manager.py
```diff
@@ -51,10 +51,11 @@ def most_recent(self):
         excluded_fields = self.get_excluded_fields()
 
         for field in self.instance._meta.fields:
-            if isinstance(field, models.ForeignKey):
-                tmp.append(field.name + "_id")
-            elif field.name not in excluded_fields:
-                tmp.append(field.name)
+            if field.name not in excluded_fields:
```
rossmechanic: I might just change this so you don't nest the `if` statement.

Try:
```
if field.name in excluded_fields:
    continue
elif isinstance(field, models.ForeignKey):
    tmp.append(field.name + "_id")
else:
    tmp.append(field.name)
```

Alig1493: Have you made the changes? If not maybe I can just do them and push it for you and it stays there?

rossmechanic: Yea can you do that? Then I’ll merge

PR #478, opened 2018-11-07
Add check for _history_excluded_fields

chaws:
## Description
Add `_history_excluded_fields` in HistoryManager  when the model belongs to Django's `__fake__` module to prevent data migrations from failing.

## Motivation and Context
On a project I'm working on, we're using two migrations to get simple_history up and running: first is the auto-generated one and the second is a data migration, that does what `./manage populate_history --auto` does.

When using data migrations and creating the up and down functions to pass to RunPython, the first argument is the Django registry that contains all installed apps where I can get a temporal version of my HistoricalModel thrgough `get_model`. The problem is that Django gives a fake model, which does not contain `_history_excluded_fields` or any other attributes besides the data ones.

I did a workaround here https://github.com/Linaro/squad/pull/379/commits/3c56e6a01e79c21e96a508599ca99467ffb2f962#diff-637039b1754cb5ac4d81dc43b33d30a1R16

Perhaps adding this check would help others creating data migrations without this workaround

## How Has This Been Tested?

Just ran simple_history tests, I don't think it'll brake anything, since it'd be adding an attribute in case `self.model` doesn't have it already

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -93,6 +93,9 @@ class HistoryManager(models.Manager):
     def bulk_history_create(self, objs, batch_size=None):
         """Bulk create the history for the objects specified by objs"""
 
+        if not hasattr(self.model, '_history_excluded_fields'):
+            self.model._history_excluded_fields = []
+
         historical_instances = [
             self.model(
                 history_date=getattr(instance, '_history_date', now()),


simple_history/manager.py
```diff
@@ -93,6 +93,9 @@ def _as_of_set(self, date):
     def bulk_history_create(self, objs, batch_size=None):
         """Bulk create the history for the objects specified by objs"""
 
+        if not hasattr(self.model, '_history_excluded_fields'):
+            self.model._history_excluded_fields = []
```
kseever: Lazy-initializing this here seems like it might be the wrong place to do this. I think somewhere within `HistoricalRecords` where it creates the history instance may be better?

chaws: I think it's already done [here](https://github.com/treyhunner/django-simple-history/blob/3183611504d80860ee11f64d47bb2a8fde0406db/simple_history/models.py#L136), which is OK when retrieving the actual model. The key problem is that within a data migration, Django will work with models that are created on-the-fly due to possible changes in model classes (temporal models), thus `_history_excluded_fields` won't be available.

rossmechanic: Only worry here is that it will make this feature unavailable to anyone who has excluded fields on their model. Any thoughts to fix that?

chaws: Yeah, we had the exact same discussion! Couldn't think of anything else other than duplicating the excluded fields in the data migration (ugh). In the way Django is currently designed, it wouldn't be possible to create a data migration that would make use of excluded fields, or you think there's another way around it? I'm trying to avoid running `populate_history` from within that data migration...

rossmechanic: After spending some time reading [this](https://docs.djangoproject.com/en/2.1/topics/migrations/#historical-models), I think we're better off documenting this issue in `Common Issues` in the docs rather than putting this change here. @chaws you should be able to do this instead:

```
HistoricalModel = apps.get_model('app_name', 'HistoricalModel')
HistoricalModel._history_excluded_fields = [] # or a list of excluded fields at the time of that migration
HistoricalModel.objects.bulk_history_create(objs)
```
Ideally, `_history_excluded_fields` would be on the `Meta` class of the Historical class, so that it could be serialized in migrations and we could avoid this, but for now django doesn't allow us to have custom fields on the meta class (https://code.djangoproject.com/ticket/5793). I'd say this makes for a pretty good argument for that ticket being reopened though.

Anyway, unless convinced otherwise, I'm going to close this PR. If you'd like, I'd love to see another PR with documentation about how to handle this issue in the Common Issues section of the docs.

chaws: That's the exact workaround I did here: https://github.com/Linaro/squad/commit/3c56e6a01e79c21e96a508599ca99467ffb2f962#diff-637039b1754cb5ac4d81dc43b33d30a1R16

I have to agree with you. Having this hardcoded in a migration would also be affected in case `_history_excluded_fields` field changes in the future.

Here's a thought after reading [this](https://docs.djangoproject.com/en/2.1/topics/migrations/#historical-models):

> They will, however, have the same fields, relationships, managers (limited to those with use_in_migrations = True) and Meta options (also versioned, so they may be different from your current ones)

I notice that HistoricalRecords is shifted to HistoryManager afterwards, which if a subclass of `django.db.models.Manager`. So by definition it should be possible to have HistoryManager access on Historical models in data migrations. I'm going to give it a try on that later, if no success, I'm happy to contribute with some documentation about it.

Thanks for prompt answers =)

PR #479, opened 2018-11-08
Don't resolve relationships for history objects

RealOrangeOne:
Don't resolve relationships when calculating the changes to a model.

## Motivation and Context
This improves 2 primary things:
- Not fetching related objects results in fewer queries, meaning diffing is faster
- If an object references a foreign key, and the row on the other end doesn't exist anymore, it won't raise an exception.

## How Has This Been Tested?
Unit tests, and manual edits on another project.

## Types of changes

This is a breaking change because the values stored in `ModelChange` etc are now the IDs rather than the model instances themselves.

## Checklist:


simple_history/models.py
```diff
@@ -26,9 +27,15 @@
     pre_create_historical_record
 )
 
+import json
+
 registered_models = {}
 
 
+def model_to_dict(model):
```
kseever: There's an existing `model_to_dict` in `django.forms.models` - seems a bit odd to depend on something from `forms` here but it may be better than reimplementing? The JSON/serialize dance is a bit wonky to me.

simple_history/models.py
```diff
@@ -448,19 +455,17 @@ def diff_against(self, old_history):
...
-        ]
-        for field in self._meta.fields:
-            if field.name in instance_fields and \
-               field.name in old_instance_fields:
-                old_value = getattr(old_history, field.name, '')
-                new_value = getattr(self, field.name)
+        old_values = model_to_dict(old_history.instance)
+        current_values = model_to_dict(self.instance)
+        for field in current_values.keys():
+            if field in current_values and \
```
kseever: I realize the equivalent conditional existed in the previous implementation, but I think it's unnecessary. `field` is coming from `current_values.keys()` so this will always evaluate to `True`. Can we drop it?

simple_history/models.py
```diff
@@ -448,19 +455,17 @@ def diff_against(self, old_history):
 
         changes = []
         changed_fields = []
-        instance_fields = [field.name for field in self.instance._meta.fields]
-        old_instance_fields = [
-            field.name for field in old_history.instance._meta.fields
-        ]
-        for field in self._meta.fields:
```
kseever: I'm wondering if there's a difference between `self.instance._meta.fields` (forming `instance_fields` above) and `self._meta.fields`? If there is, I think the new code might not be iterating over the same thing.

RealOrangeOne: There is a difference. Namely in the lack of `history_*` fields, which we don't want to be diffing anyway

RealOrangeOne:
It's been on the back burner a little, but it's on my list to revisit at some point, as this is quite a nice improvement. 

I'll hopefully find some time to make the changes soon!

simple_history/models.py
RealOrangeOne: Primarily it means it's obvious when queries are done, which is useful. It looks like the one in `django.models.forms` also handles many-to-many relationships, which may cause weirdness / unnecessary queries

rossmechanic:
@RealOrangeOne thanks for updates. Looks like you just need to run `make format` to fix the `black` errors

PR #483, opened 2018-11-16
Duplicate history cleanup

### LINKED ISSUE #313: Deleting select historical records
Not sure if this is related to the situation raised in #254, but the app had a runaway save problem and created a ton of nearly duplicate historical entries which need to be cleared.

Running a routine that found each of the duplicate historical objects from `Model.history.first().instance`. And then using `.delete()` seems to delete the original record rather than the history record.

Is it possible to do this through the queryset manager?

fopina:
Management command to find (and remove) duplicate entries in history

## Description

Management command to find (and remove) duplicate entries in history `clean_duplicate_history`

## Related Issue
fixes #313 

## Motivation and Context
As discussed in #313 there are cases where we just `.save()` without checking if anything actually changed (as we didn't have the original data loaded).  
For the same reason, doing that check in `simple_history` would have a big performance impact (force a SELECT before any UPDATE) for something that is not required by everyone (as I also prefer performance over clean history).

But we can have both (performance and clean history) if we clean up the history "later" hence I made a management command like this one over an year ago and been using it in quite a few apps so far.

Now I finally decided to rewrite it according to the code style of this project and share it :)

## How Has This Been Tested?
Unit test made (and included) and manual testing using a test app (basically doing the same scripted in the unit test)
* Create instance
* Use admin to edit and save a few times (without changing data)
* Do it once again changing data
* Run command and verify that only 2 entries of history exist (creation and latest one where data actually changed)

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/management/commands/clean_duplicate_history.py
+++ b/simple_history/management/commands/clean_duplicate_history.py
@@ -28,14 +28,10 @@ class Command(populate_history.Command):
             "type",
         )
         parser.add_argument(
-            '-d', '--dry',
-            action='store_true',
-            help='Dry (test) run only, no changes'
+            "-d", "--dry", action="store_true", help="Dry (test) run only, no changes"
         )
         parser.add_argument(
-            '-m', '--minutes',
-            type=int,
-            help='Only search the last MINUTES of history'
+            "-m", "--minutes", type=int, help="Only search the last MINUTES of history"
         )
 
     def handle(self, *args, **options):
@@ -63,7 +59,7 @@ class Command(populate_history.Command):
             if self.verbosity >= 1:
                 self.stdout.write(self.COMMAND_HINT)
 
-        self._process(to_process, date_back=options['minutes'], dry_run=options['dry'])
+        self._process(to_process, date_back=options["minutes"], dry_run=options["dry"])
 
     def _process(self, to_process, date_back=None, dry_run=True):
         if date_back:
@@ -109,9 +105,9 @@ class Command(populate_history.Command):
             if extra_one:
                 entries_deleted += self._check_and_delete(f1, extra_one, dry_run)
         if self.verbosity >= 1:
-            self.stdout.write(self.DONE_CLEANING_FOR_MODEL.format(
-                model=model, count=entries_deleted
-            ))
+            self.stdout.write(
+                self.DONE_CLEANING_FOR_MODEL.format(model=model, count=entries_deleted)
+            )
 
     def _check_and_delete(self, entry1, entry2, dry_run=True):
         delta = entry1.diff_against(entry2)
--- a/simple_history/management/commands/clean_duplicate_history.py
+++ b/simple_history/management/commands/clean_duplicate_history.py
@@ -1,0 +1,122 @@
+from django.utils import timezone
+from django.db import transaction
+
+from . import populate_history
+from ... import models, utils
+from ...exceptions import NotHistoricalModelError
+
+get_model = populate_history.get_model
+
+
+class Command(populate_history.Command):
+    args = "<app.model app.model ...>"
+    help = (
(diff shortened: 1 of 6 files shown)


rossmechanic:
@fopina thanks for this! Do you mind trying to hit complete coverage on this PR so that coverage doesn't drop below 97%? Also, if you could add a confirmation prompt in the management command that would also be great. I think @kseever is going to look into using https://github.com/romgar/django-dirtyfields so that duplicates aren't saved in the first place, but this management command will be useful for purging all duplicates. Thanks again!

fopina:
I've used dirtyfields before and I guess it should work well. But I don't really mind the duplicate history as long as it is eventually cleaned up, plus, I'm using version 1.9.0 in my apps and `.diff_against` wasn't there, but I had very similar function. Now with that helper it becomes easier for a future PR that highlights the actual changed fields in the admin interface :)

fopina:
Not really sure what to do about that failing check, let me know if something needs to be done :)

simple_history/management/commands/clean_duplicate_history.py
```diff
@@ -0,0 +1,118 @@
+from django.utils import timezone
+from django.db import transaction
+
+from . import populate_history
+from ... import models, utils
+from ...exceptions import NotHistoricalModelError
+
+get_model = populate_history.get_model
```
rossmechanic: just call `populate_history.get_model` when you need it

simple_history/management/commands/clean_duplicate_history.py
```diff
@@ -0,0 +1,118 @@
...
+
+        else:
+            if self.verbosity >= 1:
+                self.stdout.write(self.COMMAND_HINT)
+
+        self._process(to_process, date_back=options["minutes"], dry_run=options["dry"])
+
+    def _process(self, to_process, date_back=None, dry_run=True):
+        if date_back:
+            stop_date = timezone.now() - timezone.timedelta(minutes=date_back)
```
rossmechanic: Should timezone.now() be specified on initialization rather than at this point? Maybe move it to the top of `handle`?

rossmechanic:
Could you add some documentation, add your name to AUTHORS.rst and add the change to CHANGES.rst?

rossmechanic:
@fopina you still working on this?

fopina:
Haven’t been able to come back to it yet, will do

simple_history/management/commands/clean_duplicate_history.py
fopina: that point is actually called only once per execution but I've moved it to the top of `handle`

fopina: just undid this change as linting complains about `handle` becoming too complex. I think it might have been that that drove me to move it to `_process` in first place, can't remember...

rossmechanic:
Mind trying to do a bit of refactoring to solve the codeclimate issues?

fopina:
I'm afraid I can't @rossmechanic 
Although I understand/agree with flake8 rules for function/line complexity, codeclimate issues raised surpass me.. apart from the last one that I assume I can just make it an `elif`, I'm not sure how to fix the others.
Some I don't even understand, others I can't find a solution for them that won't trigger other issues or make code look "stupid" (as in moving 2 lines to separate functions used only in one place but in order to reset `complexity counters`), that's not really making code easier to understand as it means jumping back and forward between functions, imo..

fopina:
Gave it a try though, 2 down :)

rossmechanic:
Thanks for tryin @fopina . Rebasing should get rid of your Radon issues. I would definitely refactor the identical code block into a `utils.py` file. Make those changes and I'll merge with the codeclimate issues. Thanks!

fopina:
There you go!

rossmechanic:
Looks great! Thanks!

PR #490, opened 2018-12-02
Support passing a callable in "custom_model_name" feature

uhurusurfa:
## Description
Added ability to pass a callable in the "custom_model_name" feature. 

## Related Issue
https://github.com/treyhunner/django-simple-history/issues/489

## Motivation and Context
The table names become quite long when prepending "Historical" to all base class names to produce the Historical model class name and would like to be able to override the default of using "Historical".
This feature would also allow model names to be in the form

    [appName]_[modelName]

where the history models are created in a dedicated app using the "app" option thus making name much more aligned with the base table it is tracking

## How Has This Been Tested?
Additional unit tests were added to the original one testing this feature where it only supported passing a string. The feature now has 4 tests.
The enhancement does not affect any functional code and existing tests for the rest of the application functionality all passed.

## Types of changes
- [ x] New feature (non-breaking change which adds functionality)

## Checklist:
- [ x] I have run the `make format` command to format my code
- [ x] I have updated the documentation accordingly.
- [ x] I have read the **CONTRIBUTING** document.
- [x ] I have added tests to cover my changes.
- [ x] I have added my name and/or github handle to `AUTHORS.rst`
- [ x] I have added my change to `CHANGES.rst`
- [ x] All new and existing tests passed.


uhurusurfa:
@rossmechanic  - no idea how to sort out the codeclimate and codecov issues.

uhurusurfa:
@rossmechanic 
I have extended the number of tests but code coverage has not changed.
I have refactored but it seems that what is now to me a very simple method is still too complex and loathe to break it down more given how small it is already.

rossmechanic:
Hey @uhurusurfa thanks for this! Don't worry about CodeClimate for now. We just enabled it in order to replace landscape.io (which seems to have stopped being maintained), but haven't figured out exactly which parts we want enabled. I'll take a look at this PR later today.

docs/advanced.rst
```diff
@@ -562,3 +562,79 @@ or the record invocation: ``app="SomeAppName"``.
...
+Custom History Model Name
+-------------------------
+
+By default the history model names are the base model name prefixed with "Historical".
+So given a model named ``Poll``, the resultant history class will be named ``HistoricalPoll``.
+If the app name of the ``Poll`` model is `election` then the history table name would be ``election_historicalpoll``.
+
+This feature allows a different and/or shorter model name and would even allow model names to be in the standard Django form
+
+    [appName]_[modelName]
```
rossmechanic: Is this the standard Django form? Can't think of when models are referenced using their app name unless using the `get_model` function

docs/advanced.rst
```diff
@@ -562,3 +562,79 @@ or the record invocation: ``app="SomeAppName"``.
...
+.. code-block:: python
+
+    class Poll(models.Model):
+        question = models.CharField(max_length=200)
+        history = HistoricalRecords(custom_model_name=lambda x:f'Audit{x}')
+
+    class Opinion(models.Model):
+        opinion = models.CharField(max_length=2000)
+
+    register(Opinion, custom_model_name_prefix=lambda x:f'Audit{x}')
```
rossmechanic:
```suggestion
    register(Opinion, custom_model_name=lambda x:f'Audit{x}')
```

docs/advanced.rst
```diff
@@ -562,3 +562,79 @@ or the record invocation: ``app="SomeAppName"``.
...
+If the app the models are defined in is `yoda` then the corresponding history table names would be `yoda_auditpoll` and `yoda_auditopinion`
+ 
+You could also use a base class for your models and declare the history instantiation in the base class.
+All models that inherit the base model would automatically get this custom name.
+Using this mechanism requires using the `inherit` option.
+
+.. code-block:: python
+
+    class BaseModel(models.Model):
+        history = HistoricalRecords(inherit=True, app="yoda", custom_model_name_prefix=lambda x:f'Audit{x}')
```
rossmechanic:
```suggestion
        history = HistoricalRecords(inherit=True, app="yoda", custom_model_name=lambda x:f'Audit{x}')
```

docs/advanced.rst
```diff
@@ -562,3 +562,79 @@ or the record invocation: ``app="SomeAppName"``.
...
+        class Meta:
+            abstract = True
+
+    class Poll(BaseModel):
+        question = models.CharField(max_length=200)
+
+    class Opinion(BaseModel):
+        opinion = models.CharField(max_length=2000)
+
+The resulting history class names would be `AuditPoll` and `AuditOpinion`.
```
rossmechanic: Not sure you need this section on inheritance, but if you want to keep it, I think you'd be better off creating a subclass of `HistoricalRecords`. For example:
```
class NewHistoricalRecords(HistoricalRecords):
    def __init__(*args, **kwargs):
        kwargs['custom_model_name'] = lambda x: f'Audit{x}'
        kwargs['app'] = 'yoga'
        super().__init__(*args, **kwargs)
```

docs/advanced.rst
```diff
@@ -562,3 +562,79 @@ or the record invocation: ``app="SomeAppName"``.
...
+The history model tables will be named `yoda_auditpoll` and `yoda_auditopinion`.
+
+If the instantiation in the above example used `lambda x:f'{x}'` for the `custom_model_name` option
+  then the resulting history class names would be `HistoricalPoll` and `HistoricalOpinion`.
```
rossmechanic: Why doesn't this change the class names?

rossmechanic: Definitely want to keep the classnames consistent here. Otherwise I would focus strictly on table names

docs/advanced.rst
```diff
@@ -562,3 +562,79 @@ or the record invocation: ``app="SomeAppName"``.
...
+  and the corresponding history table names would be `yoda_poll` and `yoda_opinion`
+ 
+ NOTE: the class name of the historical class in this situation would be `HistoricalPoll` and NOT just `Poll`
+       to ensure it is different to the tracked class since using the `inherit` option that is necessary when using
+       an abstract class forces the historical model to be created in the same module as the model class the history model
+       is created from and would otherwise have the identical name.
+ 
+IMPORTANT: Setting `custom_model_name` to `lambda x:f'{x}'` requires that the history tables
```
rossmechanic: Why?

simple_history/models.py
```diff
@@ -133,6 +133,34 @@ def finalize(self, sender, **kwargs):
         setattr(sender, self.manager_name, descriptor)
         sender._meta.simple_history_manager_attribute = self.manager_name
 
+    def check_custom_model_name_unique(self, model):
+        # if the custom model name is the same as the original model name,
+        # make sure that the app names are different
+        try:
```
rossmechanic: Weird try block. Why isn't this just 
```
if not self.app or self.app == model._meta.app_label:
    raise ValueError( "The 'custom_model_name'....)
```

uhurusurfa: I agree but saw it done that way in the __init__ method and assumed it might be a convention used in your app so did it the same ... will change.

docs/advanced.rst
uhurusurfa: The resulting history class name for history on model "models.Poll" would be "models.Poll" because of  "inherit=True". Though clearly it is a diffferent instance of the class I was not sure if that could somehow cause issues so went the safe route and ensured the history class is different. If you are happy with them being the same I can change it.

docs/advanced.rst
uhurusurfa: Same as below comment on history model having identical name ot the model it is tracking ... can change this if you are happy it will not cause issues.

docs/advanced.rst
uhurusurfa: Would that not require you put the instantiation in all the concrete class then? So would end up like this:

class NewHistoricalRecords(HistoricalRecords):
    def __init__(*args, **kwargs):
        kwargs['custom_model_name'] = lambda x: f'Audit{x}'
        kwargs['app'] = 'yoga'
        super().__init__(*args, **kwargs)

    class Poll(BaseModel):
        question = models.CharField(max_length=200)
        history = NewHistoricalRecords()

    class Opinion(BaseModel):
        opinion = models.CharField(max_length=2000)
       history = NewHistoricalRecords()

docs/advanced.rst
uhurusurfa: Changed

docs/advanced.rst
uhurusurfa: Changed

docs/advanced.rst
uhurusurfa: Yes was not well written - Have reworded it

docs/advanced.rst
rossmechanic: Thought the goal was to have a consistent naming convention with the same app for history, not just inheritance

docs/advanced.rst
rossmechanic: Ah got it. What's the functionality if the custom_model_name is a string that's the same as the model_name? Does it error out? Don't think you need to note this, just have it throw an error in that case.

docs/advanced.rst
rossmechanic: I think we shouldn't allow this functionality. The model name has to be different

docs/advanced.rst
uhurusurfa: It only generates a warning if the "custom_model_name" is a string with the same name as the class it will track (same 'app' and inherit=False):
`RuntimeWarning: Model 'xxx.Yyy' was already registered. Reloading models is not advised as it can lead to inconsistencies, most notably with related models`

So is not caught within history app. Want me to check for it in the same way as done for callable mechanism?

rossmechanic: It doesn't generate the warning if the `custom_model_name` is a callable?

uhurusurfa: No because the code checks for this and prepends "Historical" to the class name:
`    def get_history_model_name(self, model, attrs):
        if self.custom_model_name:
            if callable(self.custom_model_name):
                name = self.custom_model_name(model._meta.object_name)
                if not name.lower() == model._meta.object_name.lower():
                    return name
                self.check_custom_model_name_unique(model)
                # Override table_name and make class name use the standard option
                # because otherwise it creates the same class as the concrete class
                attrs["Meta"].db_table = self.app + "_" + name.lower()
            else:
                # Treat as a simple string
                return self.custom_model_name
        return "Historical{}".format(model._meta.object_name)
`

uhurusurfa: Actually it only prepends "Historical" if the "app" names are different. 
If the app names are the same it  raises and error ... so can make it do the same for string based "custom_model_name"

docs/advanced.rst
uhurusurfa: Resolving this as it is the same as issue as the thread below

docs/advanced.rst
uhurusurfa: From my own point of view the goal was to be able to keep the table name short and in the same format as the tracked table but in a different app. The classname for the historical model is not really of interest to me.

rossmechanic: Then this seems like a roundabout way of getting there. If we're going to use a custom_model_name, I'd prefer that it's used for the actual model name as opposed for the table name. Maybe you're better off allowing the `table_name` parameter be a callable?

uhurusurfa: Agreed but it seemed like the right way to do it since the classname by default defines the table name so figured it would be best to try to "stay within the reservation" as it were.
I can remove setting the table name and simply throw an error if the class name for the historical model would be the same as the model it is tracking and then for that special case add the callable on "table_name" in a separate piece of work?

rossmechanic: So you'd make it such that the historical model name can be anything except for the current model name? And do a separate PR for a custom table name? That's ok with me

uhurusurfa: So shall I just add to the current PR or finish this and start another based on a separate issue I create?

rossmechanic: Let's keep them separate

rossmechanic: Also remember to rebase!

uhurusurfa:
@rossmechanic Changes made as discussed... ready for merging?

simple_history/tests/tests/test_models.py
```diff
@@ -677,19 +679,68 @@ def test_create_history_model_with_multiple_one_to_ones(self):
...
+        clazz = OverrideModelNameUsingBaseModel1
+        expected_clazz_name = "Audit{}".format(clazz.__name__)
+        self.verify_custom_model_name_feature(
+            clazz, expected_clazz_name, "tests_" + expected_clazz_name.lower()
+        )
+
+        from ..models import OverrideModelNameUsingExternalModel1
+
+        clazz = OverrideModelNameUsingExternalModel1
+        expected_clazz_name = "Audit{}".format(clazz.__name__)
```
rossmechanic: use `cls` instead of `clazz`

simple_history/tests/tests/test_models.py
```diff
@@ -677,19 +679,68 @@ def test_create_history_model_with_multiple_one_to_ones(self):
...
             )
 
+    def verify_custom_model_name_feature(
+        self, model, expected_class_name, expected_table_name
+    ):
+        history_model = model.history.model
+        self.assertEqual(history_model.__name__, expected_class_name)
+        self.assertEqual(history_model._meta.db_table, expected_table_name)
+
     def test_instantiate_history_model_with_custom_model_name(self):
```
rossmechanic: Let's split this up into a bunch of different tests? I try to avoid a bunch of tests in the same function because it becomes very difficult to pinpoint bugs when something breaks

uhurusurfa:
This is trying to cater for models registered from an external app and
would therefore have a different module path than the history being created
so even if the class names are the same they are not in the same package
which would be ok

simple_history/models.py
```diff
@@ -133,6 +133,28 @@ def finalize(self, sender, **kwargs):
...
+        # Must be trying to use a custom history model name
+        if callable(self.custom_model_name):
+            name = self.custom_model_name(model._meta.object_name)
+        else:
+            #  simple string
+            name = self.custom_model_name
+        # Desired class name cannot be same as the model it is tracking
+        if (
+            name.lower() != model._meta.object_name.lower()
+            or model.__module__ != self.module
```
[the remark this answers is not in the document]
uhurusurfa: This is trying to cater for models registered from an external app and would therefore have a different module path than the history being created so even if the class names are the same they are not in the same package which would be ok

simple_history/tests/tests/test_models.py
uhurusurfa: ok

simple_history/tests/tests/test_models.py
uhurusurfa: will break up

simple_history/models.py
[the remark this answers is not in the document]
uhurusurfa: Can change logic to:
`not (name.lower() == model._meta.object_name.lower()
            and model.__module__ == self.module)`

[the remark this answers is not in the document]
rossmechanic: I prefer that logic. However, the real issue is that `self.module` might not actually tell you the module. If you look at https://github.com/treyhunner/django-simple-history/blob/master/simple_history/tests/models.py#L405 in tests, `self.module` is actually the module of its inherited Abstract class rather than `tests.models`. I think you need to use `model.__module__ == attrs["__module__"]`, but that should be tested regardless. Ideally you'd have a test showing why the current conditional doesn't work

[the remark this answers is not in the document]
rossmechanic: @uhurusurfa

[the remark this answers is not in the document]
rossmechanic: Hey @uhurusurfa are you still working on this?

rossmechanic:
@uhurusurfa are you still working on this?

uhurusurfa:
@rossmechanic - yes ... been a long festive season. Will getr on it this weekend

rossmechanic:
@uhurusurfa any update?

uhurusurfa:
@rossmechanic - will have it reworked by end of weekend latest if not before

rossmechanic:
@uhurusurfa great! Thanks!

uhurusurfa:
@rossmechanic  - ready for review

uhurusurfa:
@rossmechanic - have rebased to latest again and ready for review

uhurusurfa:
@rossmechanic - is there any chance this will be merged soon?

rossmechanic:
Sorry. Been super busy with something but will try to take a look later this week 

Ross L. Mechanic
ross.mechanic@icloud.com | (917) 887 - 6551

docs/advanced.rst
```diff
@@ -0,0 +1,605 @@
+Advanced Usage
+==============
```
rossmechanic: What's this addition? Looks like you had an issue rebasing

uhurusurfa: @rossmechanic  - sorry about that. Have fixed.

docs/historical_model.rst
```diff
@@ -122,13 +122,42 @@ Users can specify a custom model name via the constructor on
...
+        question = models.CharField(max_length=200)
+        history = HistoricalRecords(custom_model_name=lambda x:f'Audit{x}')
+
+    class Opinion(models.Model):
+        opinion = models.CharField(max_length=2000)
+
+    register(Opinion, custom_model_name=lambda x:f'Audit{x}')
+    
+The resulting history class names would be `AuditPoll` and `AuditOpinion`.
+If the app the models are defined in is `yoda` then the corresponding history table names would be `yoda_auditpoll` and `yoda_auditopinion`
```
rossmechanic: Is this true? looking at `table_name` param and trying to figure out how this might conflict with it. What happens if I specify `table_name` and `custom_model_name`?

uhurusurfa: Specifying the table_name param overrides any custom_model_name setting since it explicitly sets the db_table meta attrib (https://github.com/treyhunner/django-simple-history/blob/372f21268963cd52f25e3d7915c48dd77c468d5a/simple_history/models.py#L218)

So the above is true only if the table_name param is not set.
The tests written for this feature verify both class name and table name change as a result of using custom_model_name ( https://github.com/treyhunner/django-simple-history/pull/490/files#diff-0772dbe490161136d6466f53ba7c2301R754)

rossmechanic:
Looks good to me. But I want @ThePumpingLemma or @kseever to take another look before merging

PR #491, opened 2018-12-03
Initial code-climate config

rossmechanic:
Working PR for setting up CodeClimate

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,3 @@
+engines:
+  pep8:
+    enabled: false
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -0,0 +1,42 @@
...
+      threshold: # language-specific defaults. an override will affect all languages.
+plugins:
+  bandit:
+    enabled: true
+  pep8:
+    enabled: false
+  radon:
+    enabled: true
+  sonar-python:
+    enabled: true
```
kseever: The [docs](https://docs.codeclimate.com/docs/sonar-python) show a `minimum_severity` argument that we may want to employ so as to not generate too much noise.

PR #507, opened 2018-12-29
create historical instance with respect to using() for multidb enviro…

erikvw:
Modify so historical record works in a multi-database setup by respecting Django's queryset chained method `using()` and the `using` keyword argument of `save()` and `delete()`.

## Description
In a multi-database environment, historical instances should be created in the same DB as the parent model instance. Currently this is not the case.

For example, if `settings` defines multiple databases:

    DATABASES={
        'default': {
            'ENGINE': 'django.db.backends.sqlite3',
        },
        'other': {
            'ENGINE': 'django.db.backends.sqlite3',
        }
    },

... the following code will create historical records in database `default` regardless of the value passed to `using()`:
    
    # book1 instance is created in `default`;
    # book1 historical instance is created in `default`.
    book1 = Book.objects.create(isbn='1-84356-028-1')

    # book2 instance is created in `other`;
    # book2 historical instance is created in `default` -- which is incorrect.
    book2 = Book.objects.using('other').create(isbn='1-84356-028-2')

... the same incorrect behavior is seen with `save()` and `delete()`:

    # book2 instance is created in `other`;
    # book2 historical instance is created in `default` -- which is incorrect.
    book2 = Book(isbn='1-84356-028-2')
    book2.save(using='other')

## Motivation and Context
We use this module in a multi-database environment and detected this problem.

## How Has This Been Tested?
Tests have been added to `test_models.py`

## Types of changes

## Checklist:

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -28,7 +28,7 @@ Authors
 - David Grochowski (`ThePumpingLemma <https://github.com/ThePumpingLemma>`_)
 - David Hite
 - Eduardo Cuducos
-- Erik van Widenfelt (@erikvw)
+- Erik van Widenfelt (`erikvw <https://github.com/erikvw>`_)
 - Filipe Pina (@fopina)
 - Florian Eßer
 - Frank Sachsenheim
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -28,6 +28,7 @@ Authors
 - David Grochowski (`ThePumpingLemma <https://github.com/ThePumpingLemma>`_)
 - David Hite
 - Eduardo Cuducos
+- Erik van Widenfelt (@erikvw)
 - Filipe Pina (@fopina)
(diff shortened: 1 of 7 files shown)


rossmechanic:
@erikvw waiting on another review here and then I'll merge. Out of curiosity, does your `AUTH_USER_MODEL`  live in multiple databases? This will fail when `using` specifies a database that doesn't have the `AUTH_USER_MODEL`, so I was thinking of adding the option of holding a custom field for history_user, where you can manually retrieve the user

rossmechanic:
@erikvw looking at this in the context of #539 and I think we made a mistake. It seems that `using` is always passed to history, regardless of whether the use specifies using in the queryset. Which means that this PR links the history to always use the same database as the base model. I'd love to find a way to make the historical model to follow whatever the router says, unless it's specifically stated somewhere that the base model's db should be the history model's db.

PR #511, opened 2019-01-11
Allow user to be tracked by explicit ID

rossmechanic:
## Description

This change allows tracking the `history_user` on a historical table with an explicit ID rather than a Django ForeignKey field. We add the following parameters to the `HistoricalRecords` constructor:
• `history_user_id_field`
• `history_user_getter`
• `history_user_setter`

`history_user_id_field` allows you to specify the type of field that your history user's primary key is (i.e., `IntegerField(null=True)` or `UUIDField(default=uuid.uuid4, null=True)`). When this parameter is specified, the `history_user` field on the historical instance is now a property rather than a ForeignKey field. There is a default getter and setter for that property that gets the user object given the `history_user_id` and sets the `history_user_id` given a user object. However, both the `history_user_getter` and `history_user_setter` can be overwritten (i.e. if you wanted to retrieve a user object using an API call with the ID). 

## Motivation and Context

In #371, we allowed dsh to handle users that weren't necessarily django's `AUTH_USER_MODEL`. However, as it stands currently, `history_user` is a django `ForeignKey` field, meaning that `history_user` has to be tracked by a Django model and has to be in the same database as our historical model (see: https://docs.djangoproject.com/en/2.1/topics/db/multi-db/#limitations-of-multiple-databases). 

Thus, by allowing developers to have the option of foregoing the ForeignKey field for an IntegerField/UUIDField with method for retrieving the user object and setting the ID, we allow django-simple-history to be used in applications where the user model might live in a separate database or may be only accessible by an API call. 

## How Has This Been Tested?

Tested setting and getting a history user that lives in a separate database than the historical table. 

## Types of changes

## Checklist:


simple_history/models.py
```diff
@@ -233,13 +252,64 @@ def copy_fields(self, model):
...
+            # Use text field with no max length, not enforced by DB anyways
+            history_change_reason_field = models.TextField(null=True)
+        else:
+            # Current default, with max length
+            history_change_reason_field = models.CharField(max_length=100, null=True)
+
+        return history_change_reason_field
+
+    @property
+    def _history_id_field(self):
```
kseever: Hmm why are these two property methods needed in this PR? Doesn't seem related to the user changes.

kseever: Ah it looks to be general refactoring from below. Definitely into pulling the logic into their own methods, but I'm not as big a fan of `@property`. Unless there's a compelling reason to do so, my preference would be to invoke them as regular methods.

simple_history/models.py
```diff
@@ -55,10 +67,13 @@ def __init__(
         history_id_field=None,
         history_change_reason_field=None,
         user_model=None,
-        get_user=default_get_user,
+        get_user=_default_get_user,
         cascade_delete_history=False,
         custom_model_name=None,
         app=None,
+        history_user_id_field=None,
```
kseever: What is the type of this arg? A classname?

simple_history/models.py
rossmechanic: Was trying to mimic how I’ve seen similar things done in Django source code. Don’t really have a preference though, so would be happy to change these into ‘get’ methods

simple_history/models.py
rossmechanic: Field instance. I.e models.IntegerField(null=True)

simple_history/models.py
```diff
@@ -55,10 +67,13 @@ def __init__(
...
+        history_user_getter=_history_user_getter,
+        history_user_setter=_history_user_setter,
```
kseever: This `__init__` is getting pretty unwieldy. Can we coalesce these arguments better? Maybe a `FieldOverrides` kind of object to group those together.

rossmechanic: Decided to continue adding params for version 2.7. Will open an issue for figuring out a better way to handle and will deprecate before removing in 3

simple_history/models.py
```diff
@@ -295,39 +366,10 @@ def get_prev_record(self):
...
-            # Use text field with no max length, not enforced by DB anyways
-            history_change_reason_field = models.TextField(null=True)
-        else:
-            # Current default, with max length
-            history_change_reason_field = models.CharField(max_length=100, null=True)
-
-        return {
-            "history_id": history_id_field,
+        extra_fields = {
+            "history_id": self._get_history_id_field(),
```
kseever: 👍

simple_history/models.py
```diff
@@ -295,39 +366,10 @@ def get_prev_record(self):
...
             "history_date": models.DateTimeField(),
-            "history_change_reason": history_change_reason_field,
-            "history_user": models.ForeignKey(
-                user_model,
-                null=True,
-                related_name=self.user_related_name,
-                on_delete=models.SET_NULL,
-            ),
+            "history_change_reason": self._get_history_change_reason_field(),
```
kseever: 👍👍

simple_history/models.py
```diff
@@ -345,6 +387,10 @@ def get_prev_record(self):
             ),
         }
 
+        extra_fields.update(self._get_history_user_fields())
```
kseever: 👍👍👍

docs/user_tracking.rst
```diff
@@ -75,8 +83,65 @@ This is very helpful when using ``register``:
...
+   is in a separate database from the user table.
+
+2. The user model that you want to use for ``history_user`` does not live within the
+   Django app, but is only accessible elsewhere (i.e. through an API call).
+
+There are three parameters to ``HistoricalRecords`` or ``register`` that facilitate
+the ability to manually track a ``history_user``.
+
+
+:history_user_id_field: An instance of field (i.e. ``IntegerField(null=True)`` or
```
ThePumpingLemma: Can you use other types? Should be `e.g.` then?

rossmechanic: Can be whatever type you want it to be. Any unique field on the user object

simple_history/tests/admin.py
```diff
@@ -3,6 +3,7 @@
 from django.contrib import admin
 
 from simple_history.admin import SimpleHistoryAdmin
+from simple_history.tests.external.models import ExternalModelCustomUserIdFieldModel
```
kseever: lol @ this name

rossmechanic: Changing it 🙄

PR #529, opened 2019-02-16
Use model view permissions to prevent a user from "reverting" to a previous version

erikvw:
In some scenarios, it is not ideal to allow users to revert a model instance to a previous version. 
Django 2.1 introduced view permissions which, if integrated into DSH, would give an administrator the ability to allow a user to add/change a model but not revert to a previous instance through the models history.

## Description

The ``SimpleHistoryAdmin`` reviews the historical model instance permissions instead of just the model instance by overriding the has_xxxx_permission methods in ModelAdmin.

* permissions granted for the historical model cannot exceed those of the model. That is, granting "view" and "change" permissions for the historical model but only "view" permission for the model means the user has "view" permission only.

* the [Revert] button is changed to a [Close] button if the user only has "view_xxxx_permission" 

* For Django < 2.1, the view option is not relevant and DSH's behavior is unaffected

* other text on the changeless and form is hidden if the user has only "view" permission

## How Has This Been Tested?
mostly, still adding some tests ...

## Types of changes

## Checklist:

- [ ] I have run the `make format` command to format my code

### DIFF
--- a/simple_history/admin.py
+++ b/simple_history/admin.py
@@ -15,12 +15,51 @@ from django.utils.html import mark_safe
 from django.utils.text import capfirst
 from django.utils.translation import ugettext as _
 
-USER_NATURAL_KEY = tuple(key.lower() for key in settings.AUTH_USER_MODEL.split(".", 1))
+USER_NATURAL_KEY = tuple(key.lower()
+                         for key in settings.AUTH_USER_MODEL.split(".", 1))
 
 SIMPLE_HISTORY_EDIT = getattr(settings, "SIMPLE_HISTORY_EDIT", False)
 
 
-class SimpleHistoryAdmin(admin.ModelAdmin):
+class HistoricalModelPermissionsAdminMixin:
+
+    def get_historical_permission_codename(self, action, opts):
+        """
+        Return the codename of the permission for the specified action.
+        """
+        return '%s_historical%s' % (action, opts.model_name)
+
+    def has_add_permission(self, request):
+        opts = self.opts
+        codename = self.get_historical_permission_codename('add', opts)
+        return request.user.has_perm("%s.%s" % (opts.app_label, codename))
+
+    def has_change_permission(self, request, obj=None):
+        opts = self.opts
+        codename = self.get_historical_permission_codename('change', opts)
+        return request.user.has_perm("%s.%s" % (opts.app_label, codename))
+
+    def has_delete_permission(self, request, obj=None):
+        opts = self.opts
+        codename = self.get_historical_permission_codename('delete', opts)
+        return request.user.has_perm("%s.%s" % (opts.app_label, codename))
+
+    def has_view_permission(self, request, obj=None):
+        opts = self.opts
+        codename_view = self.get_historical_permission_codename('view', opts)
+        codename_change = self.get_historical_permission_codename(
+            'change', opts)
+        return (
+            request.user.has_perm('%s.%s' % (opts.app_label, codename_view)) or
+            request.user.has_perm('%s.%s' % (opts.app_label, codename_change))
+        )
+
+    def has_view_or_change_permission(self, request, obj=None):
+        return (self.has_view_permission(request, obj)
+                or self.has_change_permission(request, obj))
+
+
+class SimpleHistoryAdmin(HistoricalModelPermissionsAdminMixin, admin.ModelAdmin):
     object_history_template = "simple_history/object_history.html"
     object_history_form_template = "simple_history/object_history_form.html"
 
@@ -50,7 +89,8 @@ class SimpleHistoryAdmin(admin.ModelAdmin):
         object_id = unquote(object_id)
         action_list = history.filter(**{pk_name: object_id})
         if not isinstance(history.model.history_user, property):
-            # Only select_related when history_user is a ForeignKey (not a property)
+            # Only select_related when history_user is a ForeignKey (not a
+            # property)
             action_list = action_list.select_related("history_user")
         history_list_display = getattr(self, "history_list_display", [])
         # If no history was found, see whether this object even exists.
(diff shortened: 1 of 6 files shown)


erikvw:
Additionally / TODO:

1. by default, nothing should change.

2. setting SIMPLE_HISTORY_REVERT_ENABLED emulates view perms only on historical objects

3. a settings attribute should activate "view permissions" and the develop / manager would then need to explicitly add/remove historical model permissions -- ideally through a user GROUP

rossmechanic:
Thanks for this @erikvw ! Have been busy but will take a look by end of week

erikvw:
that's OK @rossmechanic, i am still working through a few things. Hope to be done before the end of the week.

erikvw:
ok, @rossmechanic. that should do it for now. Note also, I ended up refactoring the ``history_view`` to get past codeclimate's complexity scoring.

rossmechanic:
@erikvw great! I'll try to comments on this this weekend.

erikvw:
Hi @rossmechanic, anything I can do to help move this along?

rossmechanic:
Hi @erikvw. sorry for the delay. Just trying to get a sense of overall approach before looking at the code. If I understand correctly
# State where `SIMPLE_HISTORY_PERMISSIONS_ENABLED=True`
• honor superuser privileges
Possible permissions:
1. view_model: can _only_ view model in admin, not history
2. change_model: can view model and change model, but not view history
3. view_historymodel: can view history model, but not model, and cannot revert history
4. change_historymodel: can view history model and revert history, but not view model or change model

# State where `SIMPLE_HISTORY_PERMISSIONS_ENABLED=False`
• honor superuser privileges;
Possible permissions:
1. view_model: can _only_ view model in admin, not history
2. change_model: can view model and change model, and view and revert history

Then, if `SIMPLE_HISTORY_REVERT_DISABLED=True` only superusers can revert history, correct? 

Just clarifying before I review the code, so that we have the same understanding.

Also, just want to let you know that at first glance this is a great, thorough PR. If you have any interest in being a maintainer, send me an email at ross.mechanic@icloud.com

rossmechanic:
@erikvw just checking to see if you saw my previous comment

erikvw:
hey @rossmechanic, i have some time tomorrow. will respond then ...

dwasyl:
Since this wasn't fully merged in yet, I jus wanted to add in 2 cents and say that rather than just hiding the `Revert` button the 'read only' view of the historical record might want to follow the pattern that Django has for `view` permission which is to render all of the fields as if there are in the `readonly_fields` list. That way the widgets are rendered which could leave some users confused UI wise.

rossmechanic:
@erikvw I like the idea suggested by @dwasyl . Thoughts?

erikvw:
@rossmechanic  and @dwasyl ... to want to follow the Django pattern of adding the fields to the "readonly_fields" is a good instinct.  The difference, however, is that the options are no longer rendered (e.g. for a radio field, etc). Although a matter of opinion/preference, my experience has been that an auditor prefers not only to see the selected response but also to see what the available options were. So I would prefer to leave it at is. What do you think?

dwasyl:
@erikvw @rossmechanic For what it's worth, I understand that and actually find it to be a problem with history records (since many of mine link to other models via ForeignKey or ManyToMany relationships which are never quite captured. The choices that you'll see in a historical record are coloured by the currently available relations or choices. Since it's not truly an accurate historical representation my thought is that displaying it as a widget makes it more confusing since an end-user could think those were the options at the time the record was made (which may or may not be true).

erikvw:
@rossmechanic @dwasyl You make a good case. Would you like me to explore this change?

rossmechanic:
@erikvw I agree with @dwasyl . If it's an easy change, feel free to add it to this, but I'm also happy to get this merged and let @dwasyl put up a follow-on PR to make the fields read only. Up to you

erikvw:
Would be nice to merge and treat @dwasyl 's suggestion as a separate PR.

simple_history/admin.py
```diff
@@ -97,45 +162,108 @@ def history_view(self, request, object_id, extra_context=None):
...
+        """
+        queryset = self.history_manager.filter(
+            **{self.model._meta.pk.attname: object_id}
+        )
+        if not isinstance(self.history_manager.model.history_user, property):
+            queryset = queryset.select_related("history_user")
+        return self.fetch_history_list_display_callables(queryset)
+
+    @property
+    def history_manager(self):
```
rossmechanic: We have a util that already does this. but fine with this for now so we can get this in

simple_history/templates/simple_history/object_history_form.html
```diff
@@ -5,11 +5,15 @@
...
-    <a href="{% url "admin:app_list" app_label %}">{{app_label|capfirst|escape}}</a> &rsaquo;
-    <a href="{{changelist_url}}">{{opts.verbose_name_plural|capfirst}}</a> &rsaquo;
-    <a href="{{change_url}}">{{original|truncatewords:"18"}}</a> &rsaquo;
+    <a href="{% url "admin:app_list" app_label %}">{{ app_label|capfirst|escape }}</a> &rsaquo;
+    <a href="{{changelist_url}}">{{ opts.verbose_name_plural|capfirst }}</a> &rsaquo;
+    <a href="{{change_url}}">{{ original|truncatewords:"18" }}</a> &rsaquo;
     <a href="../">{% trans "History" %}</a> &rsaquo;
-    {% blocktrans with original_opts.verbose_name as verbose_name %}Revert {{verbose_name}}{% endblocktrans %}
+	{% if has_revert_permission %}
+    	{% blocktrans with original_opts.verbose_name as verbose_name %}Revert {{ verbose_name }}{% endblocktrans %}
```
ThePumpingLemma: Indentation looks off here.

rossmechanic:
Hey @erikvw any update on this?

rossmechanic:
@erikvw lmk if you want me to take this over

erikvw:
been very busy recently, ...I'll try to look at this either today or
tomorrow. but failing that I may ask you to take over ...

rossmechanic:
Sounds good @erikvw ! Just lmk

jeking3:
This needs to be rebased.

PR #536, opened 2019-03-20
added the possibility to create a relation to the original model

nick-traeger:
## Description
i have created an optional parameter "related_name" which can be used to create another field "history_relation". This will then create a relation to the original entry.

## Related Issue
No related issue

## Motivation and Context
This should make filtering a little easier. For example, I can get all entries where a certain user has made changes.

## How Has This Been Tested?
Added some unit tests (RelatedNameTest)

## Types of changes

## Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,10 @@ Changes
 
 - remove reference to vendored library ``django.utils.six`` in favor of ``six`` (gh-526)
 
+2.7.1 (2019-03-20)
+------------------
+- Added the possibility to create a relation to the original model
+
 2.7.0 (2019-01-16)
 ------------------
 - Add support for ``using`` chained manager method and save/delete keyword argument (gh-507)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,8 +3,8 @@ Changes
 
 - remove reference to vendored library ``django.utils.six`` in favor of ``six`` (gh-526)
 
-2.7.1 (2019-03-20)
-------------------
+Unreleased
+----------
 - Added the possibility to create a relation to the original model
 
 2.7.0 (2019-01-16)
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -326,8 +326,7 @@ class HistoricalRecords(object):
             return {
                 "history_relation": models.ForeignKey(
                     model,
-                    null=True,
-                    on_delete=models.SET_NULL,
+                    on_delete=models.DO_NOTHING,
                     related_name=self.related_name,
                     db_constraint=False,
                 )
@@ -454,7 +453,7 @@ class HistoricalRecords(object):
             attrs[field.attname] = getattr(instance, field.attname)
 
         relation_field = getattr(manager.model, "history_relation", None)
-        if relation_field is not None and history_type != "-":
+        if relation_field is not None:
             attrs["history_relation"] = instance
 
         history_instance = manager.model(
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -78,6 +78,7 @@ class HistoricalRecords(object):
         history_user_id_field=None,
         history_user_getter=_history_user_getter,
         history_user_setter=_history_user_setter,
+        related_name=None,
     ):
         self.user_set_verbose_name = verbose_name
         self.user_related_name = user_related_name
@@ -93,6 +94,7 @@ class HistoricalRecords(object):
         self.user_id_field = history_user_id_field
         self.user_getter = history_user_getter
         self.user_setter = history_user_setter
+        self.related_name = related_name
 
         if excluded_fields is None:
             excluded_fields = []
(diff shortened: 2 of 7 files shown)


nick-traeger:
Is everything okay with the source code and merge request? And is there a chance that the feature will be implemented?

rossmechanic:
Hey @nick-traeger. Really interesting PR. I've got a backlog of these (5 currently open) that I'm going to try to handle this weekend, but I'll take a look soon and try to get this into the next release. 

Initial thoughts: 
• Definitely want to have some sort of warning if you have a name clash (i.e. someone does `history = historical_records(related_name=history))
• Add docs

Overall, looks good tho. @ThePumpingLemma @kseever

nick-traeger:
I have implemented a check that prevents the related_name from being equal to the history manager.

nick-traeger:
I have now extended the documentation. Is there anything else I should change?

simple_history/models.py
```diff
@@ -432,6 +452,10 @@ def create_historical_record(self, instance, history_type, using=None):
         for field in self.fields_included(instance):
             attrs[field.attname] = getattr(instance, field.attname)
 
+        relation_field = getattr(manager.model, "history_relation", None)
+        if relation_field is not None:
+            attrs["history_relation"] = instance
```
rossmechanic: What happens here when the base object is deleted? I imagine something goes wrong, since you'll have a foreign key to a null object

simple_history/models.py
```diff
@@ -315,6 +317,23 @@ def _get_history_user_fields(self):
...
+            if self.manager_name == self.related_name:
+                raise exceptions.RelatedNameConflictError(
+                    "The related name must not be called like the history manager."
+                )
+            return {
+                "history_relation": models.ForeignKey(
+                    model,
+                    null=True,
+                    on_delete=models.SET_NULL,
+                    related_name=self.related_name,
```
rossmechanic: should probably add `db_constraint=False`

simple_history/models.py
nick-traeger: There was a bug which stored the id of the old instance. But now he stores None, which is no problem at all

simple_history/models.py
nick-traeger: Added

simple_history/models.py
```diff
@@ -315,6 +317,24 @@ def _get_history_user_fields(self):
...
+        if self.related_name:
+            if self.manager_name == self.related_name:
+                raise exceptions.RelatedNameConflictError(
+                    "The related name must not be called like the history manager."
+                )
+            return {
+                "history_relation": models.ForeignKey(
+                    model,
+                    null=True,
+                    on_delete=models.SET_NULL,
```
rossmechanic: actually probably want to have models.DO_NOTHING here. If you set null, and then try to undelete an object, this pattern no longer works.

nick-traeger: Accepted and changed.

simple_history/tests/tests/test_models.py
```diff
@@ -1413,3 +1416,47 @@ def test_history_user_does_not_exist(self):
...
+        )
+        self.assertEqual(
+            Street.objects.filter(history__history_user=self.user_two.pk).count(), 2
+        )
+
+    def test_name_equals_manager(self):
+        with self.assertRaises(RelatedNameConflictError):
+            register(Place, manager_name="history", related_name="history")
+
+    def test_deletion(self):
```
rossmechanic: Can you add one more test that deletes the object and then reverts it, and see if the history queries still work? Will approve after that. Thanks

nick-traeger: Done

simple_history/tests/tests/test_models.py
```diff
@@ -1413,3 +1416,61 @@ def test_history_user_does_not_exist(self):
...
+        self.two.delete()
+
+        self.assertEqual(Street.log.filter(history_relation=2).count(), 2)
+        self.assertEqual(Street.log.count(), 4)
+
+    def test_revert(self):
+        id = self.one.pk
```
rossmechanic: this test is a little confusing to me. Was hoping for something that would tell me that all of the previous historical records are still related to the reverted object, so:
```
id = self.one.id 
self.one.delete()

self.assertEqual(0, Street.objects.filter(pk=id).count())

old = Street.history.filter(id=id).first()
old.history_object.save()
self.assertEqual(4, old.history.count())

Comment (rossmechanic) [reply simple_history/tests/tests/test_models.py]: Essentially just want to make sure that the same key is reused and all the foreign keys still link to the same base object

Comment (nick-traeger) [reply simple_history/tests/tests/test_models.py]: I now test whether the object behaves as before and whether the relations still exist

File: CHANGES.rst [hunk 6]
```diff
@@ -3,6 +3,10 @@ Changes
 
 - remove reference to vendored library ``django.utils.six`` in favor of ``six`` (gh-526)
 
+2.7.1 (2019-03-20)
```
rossmechanic: The should be `Unreleased` until we release.

rossmechanic: Sorry I missed that! Last fix and then I'll approve lol

nick-traeger: Done

PR #539, opened 2019-03-26
Allow the use of multiple databases

dopatraman:
Right now, this library does not support saving historical to a database separate from its operational model. This PR allows the historical record to be saved to the same db as the historical record table.

## Description
- Save historical record to the database associated with the historical record table by passing in the db `alias` contained in the historical instance.

## Motivation and Context
As of now, the historical records are created using the database associated with the operational model being tracked, _not_ the database associated with the historical table. That means if the operational model is in a separate database, the library attempts to look up the historical table in the operational database, and subsequently fails.

## How Has This Been Tested?
So far, I have not been able to get the full test suite to run locally.

That being said, the tests involve creating a record on a second database by passing the db name into the `HistoricalRecords` constructor and then asserting its existence.

## Types of changes

## Checklist:

### DIFF
--- a/docs/multiple_dbs.rst
+++ b/docs/multiple_dbs.rst
@@ -39,3 +39,36 @@ an issue where you want to track the history on a table that lives in a separate
 database to your user model. Since Django does not support cross-database relations,
 you will have to manually track the ``history_user`` using an explicit ID. The full
 documentation on this feature is in :ref:`Manually Track User Model`.
+
+Migrating Historical Tables to Multiple Databases
+-------------------------------------------------
+If you want your historical records to live in separate databases, you will need
+to write a [database router](https://docs.djangoproject.com/en/2.1/topics/db/multi-db/#database-routers).
+The router interface requires 4 methods, but the only one that concerns us here
+is `allow_migrate`:
+```
+class DbRouter(object):
+    ...
+    def allow_migrate(self, db, app_label, model_name, **kwargs):
+        if self._is_history_db(db, app_label, model_name, kwargs):
+            return True
+        elif self._is_default_db(db, app_label, model_name, kwargs):
+            return True
+        else:
+            return False
+```
+Your `_is_history_db` method can be implemented however you like.
+
+Tracking History in a Separate Database
+---------------------------------------
+If you want to manage a historical model in separate database from its
+operational model, you may pass the `using` keyword to the `HistoricalRecords`
+constructor:
+```
+class MyModel(models.Model):
+    ...
+    history = HistoricalRecords(using='history_db')
+```
+
+As long as your historical table has been migrated to the `history_db`, you will
+be able to write to that database on every history event.
--- a/docs/multiple_dbs.rst
+++ b/docs/multiple_dbs.rst
@@ -72,3 +72,18 @@ class MyModel(models.Model):
 
 As long as your historical table has been migrated to the `history_db`, you will
 be able to write to that database on every history event.
+
+## Testing
+If you wish to test the existence of historical models in separate databases,
+you will need to add a Meta inner class:
+
+```
+class ModelWithHistoryInDifferentDb(models.Model):
+    name = models.CharField(max_length=30)
+    history = HistoricalRecords(using="other")
+
+    class Meta:
+        app_label = "external"
+```
+
+The app_label must be set to "external" for the migration to apply to the test database.
--- a/docs/multiple_dbs.rst
+++ b/docs/multiple_dbs.rst
@@ -40,50 +40,17 @@ database to your user model. Since Django does not support cross-database relati
 you will have to manually track the ``history_user`` using an explicit ID. The full
 documentation on this feature is in :ref:`Manually Track User Model`.
 
-Migrating Historical Tables to Multiple Databases
--------------------------------------------------
-If you want your historical records to live in separate databases, you will need
-to write a [database router](https://docs.djangoproject.com/en/2.1/topics/db/multi-db/#database-routers).
-The router interface requires 4 methods, but the only one that concerns us here
-is `allow_migrate`:
-```
-class DbRouter(object):
-    ...
-    def allow_migrate(self, db, app_label, model_name, **kwargs):
-        if self._is_history_db(db, app_label, model_name, kwargs):
-            return True
-        elif self._is_default_db(db, app_label, model_name, kwargs):
-            return True
-        else:
-            return False
-```
-Your `_is_history_db` method can be implemented however you like.
+Tracking History Separate from the Base Model
+---------------------------------------------
+You can choose whether or not to track models' history in the same database by
+setting the flag `use_base_model_db`.
 
-Tracking History in a Separate Database
----------------------------------------
-If you want to manage a historical model in separate database from its
-operational model, you may pass the `using` keyword to the `HistoricalRecords`
-constructor:
 ```
 class MyModel(models.Model):
     ...
-    history = HistoricalRecords(using='history_db')
+    history = HistoricalRecords(use_base_model_db=False)
 ```
 
-As long as your historical table has been migrated to the `history_db`, you will
-be able to write to that database on every history event.
-
-## Testing
-If you wish to test the existence of historical models in separate databases,
-you will need to add a Meta inner class:
-
-```
-class ModelWithHistoryInDifferentDb(models.Model):
-    name = models.CharField(max_length=30)
-    history = HistoricalRecords(using="other")
-
-    class Meta:
-        app_label = "external"
-```
-
-The app_label must be set to "external" for the migration to apply to the test database.
+If set to `True`, migrations and audit
+events will be sent to the same database as the base model. If `False`, they
+will be sent to the place specified by the database router.
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -1493,10 +1493,15 @@ class UsingSeparateDatabaseTestCase(TestCase):
         self.user = get_user_model().objects.create(
             username="username", email="username@test.com", password="top_secret"
         )
-        self.model_ = ModelWithHistoryInDifferentDb(name="test")
-
-    def tearDown(self):
-        self.model_ = None
 
     def test_using_separate_db(self):
-        self.assertEqual("other", self.model_.history.db)
+        self.model_ = ModelWithHistoryInDifferentDb(name="test")
+        self.model_.save()
+        self.assertEqual("other", self.model_.history.using('other').db)
+        self.assertEqual(1, self.model_.history.using('other').count())
+        self.assertEqual("+", self.model_.history.using('other').first().history_type)
+        self.model_.name = "test1"
+        self.model_.save()
+        self.assertEqual("other", self.model_.history.using('other').db)
+        self.assertEqual(2, self.model_.history.using('other').count())
+        self.assertEqual("~", self.model_.history.using('other').first().history_type)
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -18,7 +18,10 @@ from django.urls import reverse
 from simple_history import register
 from simple_history.exceptions import RelatedNameConflictError
 from simple_history.models import HistoricalRecords, ModelChange
-from simple_history.signals import pre_create_historical_record
+from simple_history.signals import (
+    pre_create_historical_record,
+    post_create_historical_record,
+)
 from simple_history.tests.custom_user.models import CustomUser
 from simple_history.tests.tests.utils import (
     database_router_override_settings,
@@ -62,6 +65,7 @@ from ..models import (
     HistoricalPollWithHistoricalIPAddress,
     HistoricalState,
     Library,
+    ModelWithHistoryInDifferentDb,
     MultiOneToOne,
     Person,
     Place,
@@ -1478,3 +1482,16 @@ class RelatedNameTest(TestCase):
 
         self.one = Street.objects.get(pk=id)
         self.assertEqual(self.one.history.count(), 4)
+@override_settings(**database_router_override_settings)
+class UsingSeparateDatabaseTestCase(TestCase):
+    def setUp(self):
+        self.user = get_user_model().objects.create(
+            username="username", email="username@test.com", password="top_secret"
+        )
+        self.model_ = ModelWithHistoryInDifferentDb(name="test")
+
+    def tearDown(self):
+        self.model_ = None
+
+    def test_using_separate_db(self):
+        self.assertEqual("default", self.model_.history.db)
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -1482,9 +1482,12 @@ class RelatedNameTest(TestCase):
 
         self.one = Street.objects.get(pk=id)
         self.assertEqual(self.one.history.count(), 4)
+
+
 @override_settings(**database_router_override_settings)
 class UsingSeparateDatabaseTestCase(TestCase):
     multi_db = True
+    db_name = "other"
 
     def setUp(self):
         self.user = get_user_model().objects.create(
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -1497,11 +1497,9 @@ class UsingSeparateDatabaseTestCase(TestCase):
     def test_using_separate_db(self):
         self.model_ = ModelWithHistoryInDifferentDb(name="test")
         self.model_.save()
-        self.assertEqual("other", self.model_.history.using('other').db)
-        self.assertEqual(1, self.model_.history.using('other').count())
-        self.assertEqual("+", self.model_.history.using('other').first().history_type)
+        self.assertEqual(1, self.model_.history.count())
+        self.assertEqual("+", self.model_.history.first().history_type)
         self.model_.name = "test1"
         self.model_.save()
-        self.assertEqual("other", self.model_.history.using('other').db)
-        self.assertEqual(2, self.model_.history.using('other').count())
-        self.assertEqual("~", self.model_.history.using('other').first().history_type)
+        self.assertEqual(2, self.model_.history.count())
+        self.assertEqual("~", self.model_.history.first().history_type)
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -26,6 +26,7 @@ from simple_history.tests.custom_user.models import CustomUser
 from simple_history.tests.tests.utils import (
     database_router_override_settings,
     middleware_override_settings,
+    database_router_override_settings_history_in_diff_db,
 )
 from simple_history.utils import get_history_model_for_model
 from simple_history.utils import update_change_reason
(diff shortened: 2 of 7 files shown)


rossmechanic:
@dopatraman thanks for this. Where is `instance.history.db` populated? Not entirely sure how your implementation would work, but open to having a solution here. You're saying that, at the moment, dsh supports multiple databases, but not having the base model and history model in separate databases, correct? Unsure why we wouldn't be able to accomplish that with routers.

dopatraman:
@rossmechanic no prob, happy to contribute!

> You're saying that, at the moment, dsh supports multiple databases, but not having the base model and history model in separate databases, correct?

Correct. Using a router crossed my mind, but it looks like argument passed to `post_save` is ultimately the decider of where the record goes. I tried to dig for the source of the `using` argument (declared in `models.py`, line 416) -- what I found was that it is coming from somewhere in Django core.

I found that if a database router sends a historical table to a separate database, the `db` property of the instance's `HistoryManager` was set to that separate database. If I had to guess, I'd say that the `db` property of `instance.history` is likely set when django runs the migration for the operational model.

I'm open to a better way to do this. But short of hacking Django core, this seemed like the best option.

Thanks!

rossmechanic:
To me, that's a bug. Ideally, history check for it's db in the following order:
1. history's own `using` parameter
2. Router's db for history
3. base table's db

Right now, it seems that the history table only accept its base table's db, overruling anything else. I'd want the solution to include options 1 and 2 without breaking anything for current users (we can then deprecate if necessary before DSH3). `instance.history.db` sounds too magical to me without knowing how exactly it's set. If it's set by the router, the solution looks fine to me, but i'd also like to add a `using` parameter to `HistoricalRecords` and have a proper hierarchy here. What do you think? @ThePumpingLemma @kseever

dopatraman:
@rossmechanic 

> i'd also like to add a using parameter to HistoricalRecords and have a proper hierarchy here.

I like this solution. Let's see if it's possible.

Side note, I'm unable to run the test and format tasks.

rossmechanic:
Don't think anyone should be using `post_save` as it's part of dsh's internal api. the hooks the are documented and are public api are `pre_create_historical_record` and `post_create_historical_record`, so I just think you pass the database that's actually being used for the historical record into those hooks

rossmechanic:
Why can't you test or format?

dopatraman:
@rossmechanic ok, I've updated the code.

FWIW, I'm not 100% happy with how I've written this because the root of the problem seems to be coming from Django itself. Do you know if there is a way to override whatever is passing `using` to the hook?

dopatraman:
This is my output for `make test`:
```
__________________________________________________________ summary __________________________________________________________
  py27-django111: commands succeeded
ERROR:  py34-django111: InterpreterNotFound: python3.4
ERROR:  py35-django111: InterpreterNotFound: python3.5
ERROR:  py36-django111: InterpreterNotFound: python3.6
  py37-django111: commands succeeded
ERROR:  py34-django20: InterpreterNotFound: python3.4
ERROR:  py35-django20: InterpreterNotFound: python3.5
ERROR:  py36-django20: InterpreterNotFound: python3.6
  py37-django20: commands succeeded
ERROR:  py35-django21: InterpreterNotFound: python3.5
ERROR:  py36-django21: InterpreterNotFound: python3.6
  py37-django21: commands succeeded
ERROR:  py35-djangotrunk: InterpreterNotFound: python3.5
ERROR:  py36-djangotrunk: InterpreterNotFound: python3.6
ERROR:   py37-djangotrunk: commands failed
  docs: commands succeeded
  flake8: commands succeeded
make: *** [test] Error 1
```

This strikes me as strange since, IIRC, tox should run tests across all of the missing environments. @rossmechanic

rossmechanic:
@dopatraman you need to install pyenv and then install all of the different versions of python using pyenv. Tox can't do anything if the interpreter doesn't exist locally. If that's not in the contributing docs, feel free to add it

dopatraman:
@rossmechanic great, that did the trick.

I'm in a bit of a quandary re: code climate style issues. The complaint is that a function should not have greater than X parameters. Unfortunately, we need to pass an extra argument to `HistoricalRecords` to ensure the log creation happens on the right database. There is probably a way to accomplish this by refactoring the code, but I think that's outside the purview of this pull request...

Is it possible to relax the rules regarding parameter count for this feature?

rossmechanic:
@dopatraman yep I can override the codeclimate issue to merge. DWAI. I'll look into this pr more closely this weekend and get any last feedback before merging

dopatraman:
@rossmechanic any progress on reviewing the PR?

rossmechanic:
@dopatraman it doesn't currently respect routers. Ideally we could have a router as the fallback rather than the base model's history. You see any way to do that?

dopatraman:
Not sure I understand what you mean by having the router as a fallback?

rossmechanic:
@dopatraman so ideally, we have:
1. history's own `using` parameter
2. Router's db for history
3. base table's db

Right now, this PR only respects 1 and 3, but if the router were to send these historical tables elsewhere, it wouldn't work. Not sure how to fix this though. Need to look more deeply into it

dopatraman:
Sorry, I'm not sure I follow. In theory, the user of this library would have their own router that sends the historical tables to some database. All that matters is that future records are saved to the same database, right?

@rossmechanic

rossmechanic:
Right now that router wouldn't work tho, right? Since the router_db will be overruled by the base table's db? Or if I specify in the router that all models starting with `historical` in their model name should be written to `db2` that would work?

dopatraman:
The db router would have to route the historical tables to the desired database. The router could do this any number of ways, for instance:
```
class DbRouter(object):
    ...
    def allow_migrate(self, db, app_label, model_name=None, **hints):
        # Check audit
        elif self._is_history_db(db, app_label, model_name):
            return True
        # default to True
        elif self._is_default_db(db, model_name, hints):
            return True
        else:
            return False
```

The implementation of these routing methods is up to the developer.

If your concern is about automating this type of configuration, it may worth looking into, but I wonder if it is within the scope of this pull request. Users of this library may want to organize their historical tables any number of ways -- as of now they have the freedom to write a db router class that meets their specifications without having the library impose something onto them.

rossmechanic:
My concern was that the `using` parameter that the base table passes to the historical table overrides the router. Such that historical tables will always either follow their own `using` parameter or the `using` parameter of the base table, but never the router itself. Is that not what you saw when you first created this PR?

dopatraman:
I see... to my knowledge the historical tables were never using the `using` parameter of the router. I'm not actually sure where the router's `using` parameter enters the scene, if at all. To my knowledge the only aspect of the router's configuration that is present on the historical model is hidden in a `_db` attribute.

The reason for this pull request _was_ to force an override of the default `using` parameter. I was under the impression that should a developer choose to supply the `using` parameter to the `HistoricalRecords` constructor, their intention would be to explicitly override the default (I was in this situation before making the PR).

rossmechanic:
There is no `using` parameter of the router. But if the historical model doesn't have a `using` parameter, it should rely on the router. The router currently cannot pick what database the historical tables are written to, which I know is outside of the scope of this PR, but touches same part of code and was hoping that we could role that into this. Not exactly sure how we would do that though, so I may accept this as is. Just need to spend some time looking into it to make sure we can't solve both problems at the same time.

rossmechanic:
Regardless, can you add docs and tests?

dopatraman:
re: docs and tests, will do.

re: router fall back I see what you're saying now. In theory we could capture the router's context when the migrations are run, and then compare the context to the inputs passed to `create_historical_model`. So, if a user creates a record on a table associated with another database but does not explicitly pass in the `using` param, we could fall back to the router context.

I'd be in favor of merging this in first and attempting the latter afterwards. I'd be happy to help build it out 😄

rossmechanic:
Great. Add the docs and tests and I'll do another pass through. I agree it should be another PR, but didn't want to do anything here that made that second PR more difficult to implement. Thanks for your help!

rossmechanic:
Yea, happy with this PR. Add docs and tests and I'll merge 👍

dopatraman:
re: tests, I'm not able  to find where we run migrations on the test databases. The `TestDbRouter` does not appear to be used by anything...

rossmechanic:
Hey @dopatraman. Docs look good. Check out these tests: https://github.com/treyhunner/django-simple-history/blob/9deb217ca38e3da324e7cfdb2d72f9e783f3d52d/simple_history/tests/tests/test_models.py#L1351-L1415 for a use of that router

dopatraman:
@rossmechanic 

So, I've added the database settings override decorator:
```
@override_settings(**database_router_override_settings)
class UsingSeparateDatabaseTestCase(TestCase):
```

and also declared my model as a part of the 'other' database:
```
class ModelWithHistoryInDifferentDb(models.Model):
    name = models.CharField(max_length=30)
    history = HistoricalRecords(using="other")
```

But still, it looks like the test db router is not a part of the call chain. Specifically, I'm not able to run the `allow_migrate()` method:

```
def allow_migrate(self, db, app_label, model_name=None, **hints):
        print('db:', db)
        print('model_name:', model_name)
```

Thoughts on how to proceed?

dopatraman:
@rossmechanic 

Given that this changeset is rather simple and I cannot get the test runner to work properly, can we merge the PR in?

rossmechanic:
Would really prefer to have it tested. My guess is you likely have to set the test case to `multi_db=True`

dopatraman:
@rossmechanic 

I agree with you and would also prefer to have tests. As far as I can tell I've followed the conventions in place to run tests across multiple databases, but can't get the tests to work as expected. I'm happy to work with you to get the tests where they need to be, but right now I'm not sure how to proceed.

I'm hoping you or someone else on the project can offer some advice?

Here's what I've done so far:
- Apply the `override_settings` decorator to the database router settings
- Set `multi_db=True` on the TestCase class
- Set `db_name='other'` on the TestCase class
- Create a model on the `other` database

These steps don't appear to actually create a model on the `other` database. Even more perplexing is that it does not seem like the `TestDbRouter` methods are getting called at all.

dopatraman:
@rossmechanic thoughts?

rossmechanic:
@dopatraman going to try to fix the tests and get back to you

dopatraman:
I've manage to fix the tests. My model needed an attribute called `app_label` set to `"external"`:
```
class ModelWithHistoryInDifferentDb(models.Model):
    name = models.CharField(max_length=30)
    history = HistoricalRecords(using="other")

    class Meta:
        app_label = "external"
```

I'll update the docs to include this.

dopatraman:
@rossmechanic I think we're set to merge... unless you have any comments?

simple_history/tests/tests/test_models.py
```diff
@@ -1478,3 +1482,21 @@ def test_revert(self):
...
+    def setUp(self):
+        self.user = get_user_model().objects.create(
+            username="username", email="username@test.com", password="top_secret"
+        )
+        self.model_ = ModelWithHistoryInDifferentDb(name="test")
+
+    def tearDown(self):
+        self.model_ = None
+
+    def test_using_separate_db(self):
```
rossmechanic: this doesn't test that `using` gets used though. Can we test for that? 
```
def test_history_saves_to_database_specified_in_using(self):
    self.model_ = ModelWithHistoryInDifferentDb(name="test")
    self.model_.save()
    from simple_history.tests.tests.utils import HistoricalModelWithHistoryInDifferentDb
    self.assertEqual(1, HistoricalModelWithHistoryInDifferentDb.objects.using('other').count())
```
Which brings up another question: if we do this, does querying break since using is only specified when writing to the database?

dopatraman: Looks like we're good on both accounts -- I've updated the tests.

rossmechanic: Oh, I figured out the problem. `ModelWithHistoryInDifferentDb` is now in `other` as well, so both tables in the same ("other") database, so these tests aren't really testing anything. I think the queries will likely fail if the base model is on a separate database. I'm thinking my preference would be to scrap the `self.using` variable, and instead have a `inherit_db` parameter that defaults to `True`. If `inherit_db` is false, then no `using` parameter is passed to `create_historical_record` and the history table relies entirely on routers. We could then add a using parameter afterward, but I think the first step needs to be allowing the routers to work correctly. Otherwise, you need to specify `using` everywhere, including the manager (so that the queryset is correct), which seems more difficult and less clean

dopatraman: Oh, i see what you mean. This is the case because I've marked `app_label = "external"`. I had to do this because the TestDbRouter was not being used.

I am currently using this branch and can verify that it is possible to store the operational model separately from the historical one and also query both. I think the root issue here is how the TestDbRouter is used.

I'm not sure what the solution involving `inherit_db` looks like.

dopatraman:
@rossmechanic 

As mentioned in the reply above, I don't think we can hook into the router without modifying django core. Since I'm on a bit of a time crunch and this changeset does not break anything that already exists, would it be possible to merge and address the `inherit_db` argument in a later refactor?

rossmechanic:
@dopatraman what I'm saying is more along the lines of this:

Pass a parameter to `HistoricalRecords` called `use_base_table_db=True`. Then add this check to both of the `create_historical_record` calls, such that the one on create would look like this:
`self.create_historical_record(instance, created and "+" or "-", using=using if self.use_base_table_db else None)`. With that, if you pass `use_base_table_db=False` to `HistoricalRecords`, then the historical model will rely on router logic rather than the base table. Would that solve your problem? If so, I'm happy to implement it right now or let you implement it.

dopatraman:
@rossmechanic 

I don't believe this would have the desired effect because `using` is coming from django core, and will only be the database of the operational table.

For illustration purposes -- suppose we have an operational model and its associated historical model. By the time `create_historical_record` is called, `using` will be set to `'default'`. Should `self.user_base_table_db` be set to `True`, the historical table will be tied to the operational table (`default`). On the off case that `self.use_base_table_db` is set to `False`, it will be even worse because `using` will be passed into the creation function as `None`.

TLDR, without the `using` parameter, the historical record can either be tied _only_ to the operational table, or not exist at all.

I don't know how we can get around forcing a table parameter on the historical record without modifying django core, which I think is outside the scope of this PR.

rossmechanic:
@dopatraman If you look at https://github.com/treyhunner/django-simple-history/blob/6b113e3e90f2405a2d9a9479bdf91afc0e695ab0/simple_history/models.py#L477 , that's where the instance is saved. If we pass `using=None` to `create_historical_record`, then that will be called with `history_instance.save(using=None)`.

Now, if you look at what the `save` call is actually doing, https://github.com/django/django/blob/master/django/db/models/base.py#L663, it's default is `using=None`, so it's fine for us to pass `using=None` to `create_historical_record`. In fact, if the `save` call is passed `using=None`, then the `save` call defers to the router to figure out which database to use:

https://github.com/django/django/blob/master/django/db/models/base.py#L700

Thus, the method described above should defer to routers. Essentially, I think we made a mistake by merging #507, but adding this flag would allow us to go back to a state where historical models can be completely controlled by routers.

dopatraman:
@rossmechanic 

I'm still not sure I understand the interface. What does `base_table` refer to here? If we say that a `HistoricalRecords` constructor has `use_base_table_db` set to True, what does that mean in the context of the history and its associated operational table?

From what I gather, if `use_base_table_db` defaults to true, we want the history to be saved to the operational database by default. If it is false, it goes to the table defined by the router. Is that correct?

rossmechanic:
@dopatraman ‘use_base_table_db’ (actually let’s call it use_base_model_db) would mean whether or not we always use the base model’s DB. When I say base model, I mean MyModel as opposed to HistoricalMyModel.  The reason we want use_base_model_db to default to True is to keep this backward compatible, wed likely change the default in the next major version

dopatraman:
@rossmechanic 

It looks like the TestDbRouter still does not migrate the models properly. I am not clear on how the migrations happen -- I am not able to hit breakpoints or log statements in the `allow_migrate` method.

dopatraman:
@rossmechanic 

I've done my best to repair the tests and refactor. Tell me what you think.

rossmechanic:
Fixed the tests for you in this branch https://github.com/treyhunner/django-simple-history/tree/using-separate-db. You can copy them over from there. You've put a lot of work into this PR so I want your version to be the one merged.

Essentially just added another router particularly for this test case. Migrations are difficult to test so we don't have to worry about them, but the tests do test that the base model and history model are read and written to the correct (separate) dbs. For the record, migrations are run before the test suite which is why they're hard to test. If you copy those tests over and update the docs, I'm happy to merge

dopatraman:
@rossmechanic 

Thank you for the assist. I like your solution -- a separate router for the override is a clean way to test. 

I've cherry-picked your changes onto this branch. It seemed quicker and I'm happy to share the credit. The tests are passing now... if you're ready, let's merge!

rossmechanic:
@dopatraman Looks good to me! Just need to update the docs and I'll merge

docs/multiple_dbs.rst
```diff
@@ -39,3 +39,37 @@ an issue where you want to track the history on a table that lives in a separate
 database to your user model. Since Django does not support cross-database relations,
 you will have to manually track the ``history_user`` using an explicit ID. The full
 documentation on this feature is in :ref:`Manually Track User Model`.
+
+Migrating Historical Tables to Multiple Databases
+-------------------------------------------------
+If you want your historical records to live in separate databases, you will need
+to write a [database router](https://docs.djangoproject.com/en/2.1/topics/db/multi-db/#database-routers).
+The router interface requires 4 methods, but the only one that concerns us here
```
rossmechanic: This isn't right. You can probably just remove this entire section on routers and only include the part on the parameter below

rossmechanic: @dopatraman

dopatraman: Technically, one would still need a database router to forward the historical table migrations right?

rossmechanic: Yep, but they'd need all 4 methods. Or at least 3 of them. If they only migrate to the other table and then can't write or read to it, there'll be issues..

rossmechanic: But that's pretty much basic Django functionality. So don't really think we need a section on it

dopatraman: Done.

simple_history/tests/tests/test_models.py
```diff
@@ -1478,3 +1483,46 @@ def test_revert(self):
...
+
+@override_settings(**database_router_override_settings_history_in_diff_db)
+class SaveHistoryInSeparateDatabaseTestCase(TestCase):
+    multi_db = True
+
+    def setUp(self):
+        self.model = ModelWithHistoryInDifferentDb.objects.create(name="test")
+
+    def test_history_model_saved_in_separate_db(self):
+        self.assertEqual(0, self.model.history.using("default").count())
```
kseever: Interesting - the history table shouldn't exist at all the default DB, right? This kind of makes it seem like it does, but that it will remain empty.

rossmechanic: So the history table exists in both here. Testing the router's ability to migrate only to one table is a lot harder (since migrations happen before any tests are run). So I figured it was enough to test that the history table is being both written and read from the `"other"` database. @kseever

dopatraman: hmm, I'm not sure I know enough about Django's ORM to tell if this is expected or not. Were you expecting an exception to be raised?

rossmechanic: @dopatraman had the router been present at the beginning of the tests (specified in the test settings as opposed to overriding the settings for these specific tests), then an error would be expected. But I think this is adequate testing without testing the migration.

simple_history/tests/tests/test_models.py
```diff
@@ -1478,3 +1483,46 @@ def test_revert(self):
...
+        self.assertEqual(1, self.model.history.count())
+        self.assertEqual(1, self.model.history.using("other").count())
+        self.assertEqual(
+            1, ModelWithHistoryInDifferentDb.objects.using("default").count()
+        )
+        self.assertEqual(1, ModelWithHistoryInDifferentDb.objects.count())
+        self.assertEqual(
+            0, ModelWithHistoryInDifferentDb.objects.using("other").count()
```
kseever: Similarly - the base model table shouldn't exist in other?

simple_history/tests/tests/test_models.py
dopatraman: Agreed, I think these tests cover our expected use. @rossmechanic if you're ready, let's merge!

simple_history/tests/tests/test_models.py
dopatraman: @rossmechanic explains this in more detail above, but the TLDR is that the migrations for tests are hard to tailor on a per-test basis. These tests ensure that history can be written to and read from separate databases than their associated operational models.

dopatraman:
@rossmechanic 

Looks like we're good to go.

dopatraman:
@rossmechanic 

Would it be possible to make a patch release with this changeset?

PR #547, opened 2019-04-15
Fix non-backward compatible issue from 2.7.0 from GH-507

rossmechanic:
## Description

#507 introduced a non-backward compatible change to DSH 2.7.0. The change affects users that were using routers beforehand to write their history models to databases other than their base models' db. #536 fixed the issue, but made it backward compatible with 2.7.0. Since then, we decided to make this next version (2.7.1) backward compatible with versions pre-2.7.0, since what was introduced in #507 was a bug. 

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/tests/models.py
+++ b/simple_history/tests/models.py
@@ -364,7 +364,19 @@ class ModelWithHistoryInDifferentApp(models.Model):
 
 class ModelWithHistoryInDifferentDb(models.Model):
     name = models.CharField(max_length=30)
-    history = HistoricalRecords(use_base_model_db=False)
+    history = HistoricalRecords()
+
+
+class ModelWithHistoryUseBaseModelDb(models.Model):
+    name = models.CharField(max_length=30)
+    history = HistoricalRecords(use_base_model_db=True)
+
+
+class ModelWithFkToModelWithHistoryUseBaseModelDb(models.Model):
+    fk = models.ForeignKey(
+        ModelWithHistoryUseBaseModelDb, on_delete=models.CASCADE, null=True
+    )
+    history = HistoricalRecords(use_base_model_db=True)
 
 
 ###############################################################################
--- a/simple_history/tests/models.py
+++ b/simple_history/tests/models.py
@@ -367,14 +367,14 @@ class ModelWithHistoryInDifferentDb(models.Model):
     history = HistoricalRecords()
 
 
-class ModelWithHistoryUseBaseModelDb(models.Model):
+class ModelWithHistoryUsingBaseModelDb(models.Model):
     name = models.CharField(max_length=30)
     history = HistoricalRecords(use_base_model_db=True)
 
 
-class ModelWithFkToModelWithHistoryUseBaseModelDb(models.Model):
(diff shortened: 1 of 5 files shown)


rossmechanic:
@erikvw I know this issue will affect you because your PR introduced the `using` functionality. Just want to make sure you're aware of it before we push 2.7.1 to pip

simple_history/tests/models.py
```diff
@@ -364,7 +364,19 @@ class ModelWithHistoryInDifferentApp(models.Model):
 
 class ModelWithHistoryInDifferentDb(models.Model):
     name = models.CharField(max_length=30)
-    history = HistoricalRecords(use_base_model_db=False)
+    history = HistoricalRecords()
+
+
+class ModelWithHistoryUseBaseModelDb(models.Model):
```
kseever: The name seems a little odd. How about `ModelWithHistoryUsingBaseModelDb`?

erikvw:
@rossmechanic Are you removing the functionality to manually select a database all together in favor of forcing the use of an explicit router?

rossmechanic:
@erikvw Nope! The issue was that #507 made it such that the base model's db (whether or not `using` was specified) had to be the same db as the historical model (overriding any routers, always). This PR makes it so that, by default, the historical model will follow whatever the router says. If you specify `use_base_model_db=True`, then the history table will be written to the same db as the base table, regardless of routers. So `use_base_model_db=True` will give you the functionality from #507

erikvw:
I agree that if a router is defined DSH should respect the router. This was a good catch. But assuming no custom router has been defined. If I save the parent with ``save(using="other")``, it will now save history with ``save(using="default")`` because ``using`` is ignored unless the new boolean is passed along with ``using``.  Am I reading this correctly?

rossmechanic:
@erikvw correct. Unfortunately I don't immediately see a clean solution for fixing that, and I'd like to get this back in a state that's backward compatible. Happy to consider options for how we might be able to handle that case afterwards. The issue was really that `using` gets passed from the base model to the historical model regardless of whether `using` is specified on the base model.

erikvw:
django makes direct reference to the router, in ``save`` for example,

``using = using or router.db_for_write(self.__class__, instance=self)``

rossmechanic:
@erikvw yea the issue is that `using` is _always_ passed to `create_historical_record` regardless of whether it's specified. So if I have 
```
MyModel.objects.create()
```
`create_historical_record` will be passed `using='default` by default

rossmechanic:
@erikvw https://docs.djangoproject.com/en/2.2/ref/signals/#post-save
`using` just specifies the database that the base model is going to use, regardless of whether the developer explicitly set `using`

rossmechanic:
You could also argue that even if I do specify `using` on the base model, I don't necessarily want the historical model to use the same database, which is why the flag makes sense IMO

erikvw:
perhaps put ``using=using if self.use_base_model_db else None`` at the top of ``create_historical_record`` instead of changing the value each time we call ``create_historical_record``.

rossmechanic:
@erikvw Ah, that is cleaner. Going to put up PR for that now

PR #555, opened 2019-04-19
Added Django 2.2 to testing matrix

rossmechanic:
Also fixed a doc issue. Doc was formatted as md instead of rst

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ install:
   - pip install -U $DJANGO
   - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then pip install black; fi
   - if [[ $TRAVIS_PYTHON_VERSION == '3.7' ]]; then pip install black; fi
-  - sqlite --version
   - pip freeze
 
 script:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -12,6 +12,7 @@ env:
   - DJANGO="Django>=1.11,<1.12"
   - DJANGO="Django>=2.0,<2.1"
   - DJANGO="Django>=2.1,<2.2"
+  - DJANGO="Django>=2.2,<2.2"
```
ThePumpingLemma: `<2.3`?

rossmechanic: Ah thank you. Actually just tried running the tests locally and they failed 😞

runtests.py
```diff
@@ -43,6 +52,7 @@
             'ENGINE': 'django.db.backends.sqlite3',
         }
     },
+    MIGRATION_MODULES=DisableMigrations(),
```
kseever: Might be a dumb question, but why not just `{}`?

simple_history/tests/tests/test_models.py
```diff
@@ -1441,6 +1445,9 @@ def test_foreign_key_to_self_using_self_str(self):
 
 @override_settings(**database_router_override_settings)
 class MultiDBExplicitHistoryUserIDTest(TestCase):
+    if django.VERSION >= (2, 2):
```
kseever: Why is this version spec different from the other two in this file? And if it's not supposed to be, could you refactor it out into a function?

runtests.py
rossmechanic: {} is the default, which would lead us back to square one with the original issue

PR #572, opened 2019-07-15
Bump version to 2.7.3

rossmechanic:
Version bump to 2.7.3 before release


CHANGES.rst
```diff
@@ -1,8 +1,8 @@
 Changes
 =======
 
-Unreleased
-----------
```
kseever: Should we just leave this section?

Issue #571, opened 2019-07-09
Historical model with "excluded_fields" causes DoesNotExist error when trying to get instance from historical record if object is deleted


yakimka:
**Describe the bug**
Historical model with "excluded_fields" causes DoesNotExist error when trying to get instance from historical record if object is deleted

**To Reproduce**
1. Create model as follows

```python
class ExampleModelWithExcludedField(models.Model):
    history = HistoricalRecords(excluded_fields=['description'])
    name = models.CharField(max_length=255)
    description = models.TextField()
    date = models.DateTimeField(auto_now_add=True)
```
2. Run `./manage.py shell`

```
>>> obj = ExampleModelWithExcludedField.objects.create(name='name', description='desc')
>>> obj_id = obj.id
>>> obj.delete()
(1, {'simple_app.ExampleModelWithExcludedField': 1})
>>> hist_obj = HistoricalExampleModelWithExcludedField.objects.filter(id=obj_id).first()
>>> hist_obj.instance
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File ".../simple_history/models.py", line 383, in get_instance
    .values(*excluded_attnames)
  File ".../django/db/models/query.py", line 379, in get
    "%s matching query does not exist." %
simple_app.models.DoesNotExist: ExampleModelWithExcludedField matching query does not exist.
```

**Expected behavior**
I expected to get instance with default values for fields that excluded.

**Environment:**
 - OS: Ubuntu 18.04, Arch linux
 - Django Simple History Version: 2.7.2
 - Django Version: 1.11.22
 - Database Version: Sqlite3, PostgreSQL 11.4

**My repo for reproduce this error**

1. Clone the repo - https://github.com/yakimka/example-simple-history-project.git
     - Database is already present.
2. Install requirements
3. Run python manage.py shell
4. Run:

```
>>> from simple_app.models import HistoricalExampleModelWithExcludedField
>>> obj_with_excluded_fields = HistoricalExampleModelWithExcludedField.objects.filter(id=1).first()
>>> obj_with_excluded_fields.instance
```

5. See error

wwarne:
Have the same issue
Django 2.2.3
Django Simple History Version: 2.7.2

rossmechanic:
This is expected behavior and can be found in this line: https://github.com/treyhunner/django-simple-history/blob/master/simple_history/models.py#L376

[...]

PR #574, opened 2019-07-16
Fixed error when trying to get instance if object is deleted #571

yakimka:
Fixed DoesNotExist error when trying to get instance if object is deleted

## Description
- changed `get_instance` method to retrieve attname and default value for field instead of only attname

## Related Issue
#571

## Motivation and Context
This change required because i think this is a bug.
It solve error when trying to get `instance`, `prev_record` or `next_record` when object is deleted

## How Has This Been Tested?
Run tests locally and updated test code

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -75,6 +75,7 @@ from ..models import (
     Poll,
     PollInfo,
     PollWithExcludeFields,
+    PollWithExcludedFieldsWithDefaults,
     PollWithExcludedFKField,
     PollWithHistoricalIPAddress,
     Province,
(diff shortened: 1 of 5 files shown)


simple_history/tests/tests/test_models.py
```diff
@@ -1185,6 +1185,18 @@ def test_restore_pollwithexclude(self):
         original = historical.instance
         self.assertEqual(original.pub_date, poll.pub_date)
 
+    def test_get_default_values_for_excluded_fields(self):
```
rossmechanic: Can you add a test where the excluded field doesn't have a default? Also, unsure whether it the excluded fields should default to the default or just be empty, but I agree that it shouldn't error out

yakimka: Yes, i will update tests. As for the defaults, I agree with you that it would be better not to pass the excluded fields - let the Django itself fill them for instance. I will update PR soon.

rossmechanic: 👍 thanks!

yakimka: Done

yakimka:
Is there anything else I should change?

PR #576, opened 2019-07-23
fix #575 use django model_to_dict to detect changes in parent models

### LINKED ISSUE #575: diff_against doesn't detect changes to parent model attributes when using inherited models
**Describe the bug**
`HistoricalChanges.diff_against` won't detect changes in inherited models (`inherit=True`).

Problem resides at the way `_model_to_dict` serializes fields.

~~~ python
def _model_to_dict(model):
    return json.loads(serialize("json", [model]))[0]["fields"]
~~~
It uses `django.core.serializers.serialize`, but as explained [here](https://docs.djangoproject.com/en/2.2/topics/serialization/#inherited-models):

> if you have a model that uses multi-table inheritance, you also need to serialize all of the base classes for the model. This is because only the fields that are locally defined on the model will be serialized.

So it ignores any fields in the parent class.

**To Reproduce**
1. Have inherited models:

~~~ python
class Place(models.Model):
    name = models.CharField(max_length=50)

class Restaurant(Place):
    serves_hot_dogs = models.BooleanField(default=False)
~~~

Then from a `Restaurant` instance change an attribute from the parent class `Place.name`

~~~ python
r = Restaurant.objects.create(name="McDonalds")
r.name = "BurgerKing"
r.save()
~~~

Now try to detect changes:

~~~ python
new_record, old_record = r.history.all()
delta = new_record.diff_against(old_record)
for change in delta.changes:
    print("{} changed from {} to {}".format(change.field, change.old, change.new))
~~~

But it won't detect the `name` field so it doesn't detect any changes and `delta.changes=[]`

 - OS: Ubuntu 18.10
 - Django Simple History Version: 2.7.2
 - Django Version: 2.2.2

marcanuy:
Use Django's `model_to_dict` as suggested at https://github.com/treyhunner/django-simple-history/issues/575#issuecomment-514047309 to detect changes in a parent model when using `inherit=True`.

## Related Issue

https://github.com/treyhunner/django-simple-history/issues/575
Related with https://github.com/treyhunner/django-simple-history/pull/479

## How Has This Been Tested?

Added test

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x ] I have run the `make format` command to format my code

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -58,6 +58,7 @@ Authors
 - Leticia Portella
 - Lucas Wiman
 - Maciej "RooTer" Urbański
+- Marcelo Canina (`marcanuy <https://github.com/marcanuy>`_)
 - Mark Davidoff
 - Martin Bachwerk
 - Marty Alchin
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changes
 Unreleased
 ----------
 - Fixed DoesNotExist error when trying to get instance if object is deleted (gh-571)
+- Use Django's `model_to_dict` instead of a custom implementation so
+  it detect changes in a parent model when using `inherit=True`
+  (gh-576)
 
 2.7.3 (2019-07-15)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,9 +4,9 @@ Changes
 Unreleased
 ----------
 - Fixed DoesNotExist error when trying to get instance if object is deleted (gh-571)
-- Use Django's `model_to_dict` instead of a custom implementation so
-  it detect changes in a parent model when using `inherit=True`
-  (gh-576)
+- Fix `model_to_dict` to detect changes in a parent model when using
+  `inherit=True` (backwards-incompatible for users who were directly
+  using previous version) (gh-576)
 
 2.7.3 (2019-07-15)
 ------------------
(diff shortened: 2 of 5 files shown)


kseever:
Hi Marcelo! Thanks for the contribution! The fix looks good to me. The only ask I have is that you add a line in CHANGES.rst about it, noting that the change is backwards-incompatible so that we include it in the next minor rev.

rossmechanic:
Hi @marcanuy any update on this?

marcanuy:
Hi @rossmechanic @kseever , I've added a line to `changes.rst` explaining this PR. Didn't mention it was backwards-incompatible as I think it doesn't break any previous functionality, does it? If so please tell me and I will include it.

kseever:
The backwards-incompatibility lies in that current callers might be subtly expecting no deltas to result from changes made in parent models. If we include this as a bugfix revision, some applications or tests might break for reasons that could be hard to understand. I think it would be good to call attention to in the changelog for folks who are looking to upgrade in case they have parent/child models that they are using the diff functionality on.

marcanuy:
@kseever Thanks for the clarification, just updated changes file

ses4j:
This change silently broke our usage of simple_history, unfortunately.  We use `editable=False` on a number of fields that we want to track with simplehistory.  Django's built-in method excludes these fields because they are not needed for forms, but we still want them in history!  Right now we're pinning to before this change.  Any ideas for a compromise?

Issue #568, opened 2019-06-14
Excluded_fields changes object name


dwasyl:
When a model has one, or multiple, fields set as `excluded_fields`, when a model is viewed on the `Change History` page instead of the string representation of the object in the `Object` column, it displays either `true` or `false`.

The problem seems to be with the name generated by `history_object` that comes from `HistoricalObjectDescriptor`, but I couldn't figure out the issues.

**To Reproduce**
1. Exclude a field from a `History()`
2. View the Change History for object of the model from above

**Expected behavior**
The string representation of the object name appears.

**Environment (please complete the following information):**
 - Django Simple History Version: 2.7.2
 - Django Version: 2.2.2

PR #585, opened 2019-09-05
Instantiate model with kwargs not args to allow for excluded fields

PeteCoward:
I just installed this project to investigate an audit log feature, and I wanted to exclude one field from my model from being tracked. I did this, but then found exceptions aplenty on accessing historical records.

Digging into the code I found that the HistoricalObjectDescriptor was instantiating its model using positional args, and my excluding the first field in my model meant it was assigning values to fields shifted by one. Which might not be noticed unless types are incompatible, like mine were as a GeoDjango spatial field will error if you attempt to assign an int to it.

My quick fix was either:
1. move my excluded field to the bottom of my model :)
2. the code change in this PR

Sorry I have no time to add tests, or the package knowledge to search for similar problems to this right now, but I figured it was better to spend 5 minutes and let you know about it :)
It looks like it might solve #568 too!

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -561,8 +561,8 @@ class HistoricalObjectDescriptor(object):
         self.fields_included = fields_included
 
     def __get__(self, instance, owner):
-        values = (getattr(instance, f.attname) for f in self.fields_included)
-        return self.model(*values)
+        values = {f.attname: getattr(instance, f.attname) for f in self.fields_included}
+        return self.model(**values)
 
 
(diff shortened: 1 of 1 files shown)


rossmechanic:
@PeteCoward thanks for this! Will try to get around to writing tests for this soon

kolotev:
The proposed fix helped me too, in case of use of "excluded_fields" property and excluded fields are declared before the one I'm interested to keep track of. Please accept the PR as soon as possible. 

Thank you.

Issue #605, opened 2019-11-19
"minutes" flag still checks historical records for records that have never changed


craigmaloney:
**Describe the bug**
Using the minutes flag on a large database will still trawl historical records that have not changed and are unlikely to change. This is fine for smaller databases but on the database that I'm using I'm finding it trawling records that are decades old and unlikely to ever change. Need better filtering on the stop date to prevent referencing these records.

**To Reproduce**
1. Create a history database on a database of 8M records
2. Run `python manage.py clean_duplicate_history -m 60 --auto`
3. Check the process to find that it is searching on unrelated records

**Expected behavior**
If we know the subset of records that we're searching on historical records it might make more sense to filter on those records from the get-go rather than iterate over each record in the database in hopes of a match. The `model.objects.all()` returned 8M records, whereas the records that had changed in that period were 1,835.

PR #604, opened 2019-11-18
Add an iterator to prevent OOM when cleaning large databases

craigmaloney:
## Description
Adds an iterator to .all() to iterate over the data rather than load it all in memory.

## Related Issue
#603 

## Motivation and Context
Prevents loading database completely into memory.

## How Has This Been Tested?
Tested in a VM with 6GB of memory that was previously (and consistently) giving an OOM error

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] I have run the `make format` command to format my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [X] I have added my name and/or github handle to `AUTHORS.rst`
- [X] I have added my change to `CHANGES.rst`
- [X] All new and existing tests passed.

Issue #603, opened 2019-11-18
OOM when using clean_duplicate_history on over 8M records


craigmaloney:
**Describe the bug**
I am getting an OOM error when running ``clean_duplicate_history`` to clean up duplicate rows for an application that I'm running.

**To Reproduce**
1. Have an existing database of over 8M records
2. Create a cloud instance of about 6GB of memory or less
3. Run ``python manage.py clean_duplicate_history --auto`` to clean up duplicates
4. Get an OOM error because the database is loaded into memory.

**Expected behavior**
Expect that the program will not OOM.

**Additional context**
This is in part because the clean_duplicate_history is using .all() without setting an iterator. The iterator will prevent loading the contents all into memory for processing.

PR #606, opened 2019-11-19
Filter the records when given a minutes argument

craigmaloney:
## Description
Filters the records so we're not having to iterate over the database if we already know what the subset of records is.

(This change incorporates the fix from #604)

## Related Issue
#605, #603

## Motivation and Context
This resolves iterating over a large database for a subset of records that we have already determined need changing.

## How Has This Been Tested?
Tested on local system.

## Types of changes

## Checklist:


rossmechanic:
mind fixing the conflicts here? @craigmaloney

craigmaloney:
> mind fixing the conflicts here? @craigmaloney

Fixed. Thanks!

simple_history/management/commands/clean_duplicate_history.py
```diff
@@ -65,11 +65,21 @@ def _process(self, to_process, date_back=None, dry_run=True):
...
+            model_query = model.objects.all()
+
+            # If we're provided a stop date take the initial hit of getting the
+            # filtered records to iterate over
+            if stop_date:
+                model_query = model_query.filter(
+                    pk__in=(m_qs.values_list(model._meta.pk.name).distinct())
+                )
+
             # it would be great if we could just iterate over the instances that
```
rossmechanic: does this change not differ from the comment here? Is this change actually faster?

rossmechanic: If so, should we remove this comment?   @craigmaloney

craigmaloney: The comment still stands if the user doesn't suggest a date to work from. If no date is specified then it'll still crawl through all of the records.

rossmechanic: But it looks like the comment mentions a given period, which would make me assume it's talking about the filtered qs

craigmaloney: I'm not sure what to do here then, because my understanding is that the comment still applies if no date is specified (eg: it doesn't just search on the changes). I left it alone because I wasn't certain if it still applied or not. My understanding is that it does.

rossmechanic: I believe the comment is describing exactly what this change is doing. Have you done any testing to see which is faster?

craigmaloney: I have tested it on my database, with my patch, which actually completed (the previous version didn't complete because it was stuck checking records from 2003).

rossmechanic: Cool. Then let's remove that comment, and then I'll merge it. Sound good?

CHANGES.rst
```diff
@@ -1,6 +1,11 @@
 Changes
 =======
 
+TBD
```
rossmechanic: We normally call this `Unreleased`

CHANGES.rst
```diff
@@ -1,6 +1,11 @@
...
+---
+
+- Add simple filtering if provided a minutes argument in `clean_duplicate_history`
```
rossmechanic: can you add `(gh-606)` to the end of this?

simple_history/management/commands/clean_duplicate_history.py
craigmaloney: Removed the comment.

CHANGES.rst
craigmaloney: OK, got it. :)

CHANGES.rst
craigmaloney: Added.

PR #625, opened 2020-02-11
Add setting to convert FileField to CharField instead of TextField

### LINKED ISSUE #623: FileField should be transformed into a CharField instead of a TextField
**Describe the bug**
`FileField` is turned into a `TextField` on history tables by the [`transform_field` function](https://github.com/treyhunner/django-simple-history/blob/68211b5dcbf79dfab71144fe1a524598b3656e23/simple_history/models.py#L546), but it should be turned into a CharField instead:

```
    elif isinstance(field, models.FileField):
        # Don't copy file, just path.
        field.__class__ = models.TextField  # should be CharField
```

Using `TextField` changes the representation of the file field in the database, because `FileField`s use the same database representation as `CharField`, not `TextField`.

For example, see Django's [postgres backend](https://github.com/django/django/blob/master/django/db/backends/postgresql/base.py#L72), where FileField is mapped to `'varchar(%(max_length)s)'` instead of `'text'`; the same is true for other backends. 

The upshot is that `FileFields` end up as `varchar`s in the primary table but as `text`s in the history table. This probably works fine for most users, since it's a more permissive field, but it's breaking a third party tool I'm using that expects the primary table and history table columns to have the same type.

**To Reproduce**
1. Create a history-tracked table with a `FileField`
2. Run `./manage.py sqlmigrate` to observe that the file field on the primary table is `varchar` but the field on the history table is `text`.

**Expected behavior**
Fields on the primary and history table should have the same type.

jcushman:
Fix #623 by adding a `SIMPLE_HISTORY_FILEFIELD_TO_CHARFIELD` setting to convert FileFields to CharFields instead of TextFields, and related documentation.

## Description
This applies the fix suggested by @rossmechanic in #623.

Any existing project that adds the new setting will have migrations created by `./manage makemigrations` to turn the history tables from `TextField` to `CharField`.

## Related Issue
#623 

## Motivation and Context
See issue

## How Has This Been Tested?
I manually tested this by running the test suite with the branch forced to `True`, and it worked.

I don't know of a way to automatically test this one! I tried adding a `TestCase` with `@override_settings(SIMPLE_HISTORY_FILEFIELD_TO_CHARFIELD=True)`, but the setting applies too late to modify the table.

I started to seriously wonder whether it makes sense to put this behind a setting at all, given the difficulty of testing -- I think it would be correct to set to True in all cases. But the fact that it will makemigrations for existing projects does make it somewhat disruptive to change as a default.

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -545,7 +545,10 @@ def transform_field(field):
 
     elif isinstance(field, models.FileField):
         # Don't copy file, just path.
-        field.__class__ = models.TextField
+        if getattr(settings, "SIMPLE_HISTORY_FILEFIELD_TO_CHARFIELD", False):
+            field.__class__ = models.CharField
+        else:
+            field.__class__ = models.TextField
 
     # Historical instance shouldn't change create/update timestamps
     field.auto_now = False
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -42,6 +42,7 @@ Authors
 - Guillermo Eijo (`guilleijo <https://github.com/guilleijo>`_)
 - Hamish Downer
 - Hanyin Zhang
+- Jack Cushman (`jcushman <https://github.com/jcushman>`_)
 - James Muranga (`jamesmura <https://github.com/jamesmura>`_)
 - James Pulec
 - Jesse Shapiro
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changes
 Unreleased
 ----------
 - Add simple filtering if provided a minutes argument in `clean_duplicate_history` (gh-606)
+- Add setting to convert `FileField` to `CharField` instead of `TextField` (gh-623)
 
 2.8.0 (2019-12-02)
 ------------------
--- a/docs/historical_model.rst
+++ b/docs/historical_model.rst
@@ -386,3 +386,16 @@ or the record invocation: ``app="SomeAppName"``.
     register(Opinion, app="SomeAppName")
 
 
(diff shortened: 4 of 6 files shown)


simple_history/models.py
```diff
@@ -545,7 +545,10 @@ def transform_field(field):
 
     elif isinstance(field, models.FileField):
         # Don't copy file, just path.
-        field.__class__ = models.TextField
+        if getattr(settings, "SIMPLE_HISTORY_FILEFIELD_TO_CHARFIELD", False):
+            field.__class__ = models.CharField
```
rossmechanic: Does this not fail if you don't specify max length?

jcushman: It works because FileFields [default to](https://github.com/django/django/blob/master/django/db/models/fields/files.py#L239) max_length of 100. 

django-simple-history's test model `FileModel` doesn't have an explicit `max_length`, for example. Here's a pdb session from inside `HistoricalRecordsTest.test_file_field` with the setting forced to True:

```
>>> FileModel.history.model._meta.get_field('file')
<django.db.models.fields.CharField: file>
>>> FileModel.history.model._meta.get_field('file').max_length
100
```

rossmechanic: Awesome. Mind just adding a test for when the setting is true?

jcushman: Done! Would have included tests before, but it took me a while to find the clever way of mocking model-setup settings in tests/models.py.

rossmechanic:
Thanks for this @jcushman !

jcushman:
Thanks for your quick review!

PR #641, opened 2020-03-21
Make history fields readonly when `SIMPLE_HISTORY_EDIT` is not set or set to `False`

partizaans:
## Description
This PR makes history form fields readonly if `SIMPLE_HISTORY_EDIT` is not set or set to `False`

## Related Issue
https://github.com/treyhunner/django-simple-history/issues/640

## How Has This Been Tested?
I manually tested history detail view once without setting `SIMPLE_HISTORY_EDIT`, once with setting it to `True` and once with setting it to `False`. I also wrote some assertions to test expected functionality. 

## Types of changes

## Checklist:

- [x] My change ~~requires~~ a change to the documentation.


tlsabara:
Hi everyone, 
I'm new here and I'm trying to use simple-history in my project. 
I was annoyed that I could edit the History before reverting. I tested the `SIMPLE_HISTORY_EDIT = False`, but for my project, it didn't work (is it not working or did I mess something up in the lib?). 
Well... anyway I worked on editing/overriding the `get_readonly_fields` function to solve my problem, and now i have two ways to solve it. But first I need to know if I'm on the right path. Can you help me please?

PS: `SIMPLE_HISTORY_EDIT = True` will create a new instance of the object in the database? (Or I again screwed up something)

If my understanding is correct I will submit a PR with my changes.

PR #655, opened 2020-04-26
Deprecate changereason

rossmechanic:
Doesn't make sense for `changeReason` to be camelCase. Replacing it with `_change_reason` and deprecating it. Will remove it completely in 3.0

### DIFF
--- a/simple_history/utils.py
+++ b/simple_history/utils.py
@@ -132,9 +132,11 @@ def get_change_reason_from_object(obj):
         return getattr(obj, "_change_reason")
 
     if hasattr(obj, "changeReason"):
-        warning_msg = "Using the attr changeReason to populate history_change_reason is" \
-                      " deprecated in 2.10.0 and will be removed in 3.0.0. Use " \
-                      "_change_reason instead. "
+        warning_msg = (
+            "Using the attr changeReason to populate history_change_reason is"
+            " deprecated in 2.10.0 and will be removed in 3.0.0. Use "
+            "_change_reason instead. "
+        )
         warnings.warn(warning_msg, DeprecationWarning)
         return getattr(obj, "changeReason")
 
--- a/simple_history/utils.py
+++ b/simple_history/utils.py
@@ -1,3 +1,5 @@
+import warnings
+
 import django
 from django.db import transaction
 from django.forms.models import model_to_dict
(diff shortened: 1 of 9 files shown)


simple_history/utils.py
```diff
@@ -123,3 +125,19 @@ def bulk_update_with_history(
             default_user=default_user,
             default_change_reason=default_change_reason,
         )
+
+
+def get_change_reason_from_object(obj):
+    if hasattr(obj, "_change_reason"):
+        return getattr(obj, "_change_reason")
```
merwok, line 132: This could be `return obj._change_reason`!
(or if this is called many times in loops, even more efficient would be `try: return obj.change_reason    except AttributeError: pass`)

simple_history/utils.py
```diff
@@ -123,3 +125,19 @@ def bulk_update_with_history(
...
+
+    if hasattr(obj, "changeReason"):
+        warning_msg = (
+            "Using the attr changeReason to populate history_change_reason is"
+            " deprecated in 2.10.0 and will be removed in 3.0.0. Use "
+            "_change_reason instead. "
+        )
+        warnings.warn(warning_msg, DeprecationWarning)
```
merwok, line 140: warnings most of the time need a `stacklevel` parameter (usually with value 2, but depends on the calling pattern) so that the message shows the calling code rather that the line with warnings.warn.

simple_history/utils.py
rossmechanic: Agree on the latter - though this will be removed in next version anyway. How could this be `return obj._change_reason` though? If that doesn't exist, it needs to check `changeReason`

simple_history/utils.py
rossmechanic: didn't know that. thanks for the feedback

simple_history/utils.py
merwok: I was saying two things at once!

1:  not reason to use getattr + string after you’ve already tested that the attribute exists
```python
if hasattr(obj, "_change_reason"):
    return obj._change_reason
```

2: if performance matters, you can avoid getting the attribute multiple times (hasattr under the covers tries to get the attr and catches AttributeError):
```python
try:
    return object._change_reason
except AttributeError:
   pass

# rest of the code that tries changeReason
```

rossmechanic: Ah yes, good points.

PR #667, opened 2020-05-23
Remove invalid reference in docs

jws:
## Description
`register` reference resolves to `Save without a historical record` which doesn't make sense. The reference looks to be leftover from past doc reorganizations. 

As there is no section which actually provides great content about `register`, better to just remove the reference completely, instead of causing confusion.

## Checklist:
- [ ] I have run the `make format` command to format my code


rossmechanic:
Thanks @jws

Issue #391, opened 2018-05-15
IntegrityError when deleting user


BojanKogoj:
I get the following exception when I try to delete a user.
```
IntegrityError at /admin/...

update or delete on table "abc_auth_user" violates foreign key constraint "measurements_histo_history_user_id_d76dbbe3_fk_abc_auth_user_id" on table "measurements_historicalusermeasurements"
DETAIL:  Key (id)=(36) is still referenced from table "measurements_historicalusermeasurements".
```

As far as I can see it should set [history_user to NULL](https://github.com/treyhunner/django-simple-history/blob/d3d8a25059690d56c2cc683e6acd19e34439ad9a/simple_history/models.py#L237) (if that's even the problem).

I'm getting this in a working code and not tests. Any known solution? It's making deleting users near impossible. I'm using version 1.9.1

rossmechanic:
Can you show me your model definition for `UserMeasurements`? There shouldn't be any db_constraints on the historical table's foreign keys because of this line: https://github.com/treyhunner/django-simple-history/blob/master/simple_history/models.py#L163

Seeing that table might help me try to help you debug.

BojanKogoj:
I removed a bunch of fields but that's it. No explicit constraints.

```python
class UserMeasurements(models.Model):

    weight = models.DecimalField(
        _('weight'), default=None, null=True, blank=True,
        max_digits=10, decimal_places=2)

    # .....

    created = models.DateTimeField(
        _('created'), auto_now_add=True)

    updated = models.DateTimeField(
        _('updated'), auto_now=True)

    history = HistoricalRecords()

    user = models.OneToOneField(
        settings.AUTH_USER_MODEL, related_name='measurements_user')

models.signals.post_save.connect(
    measurements_updated, sender=UserMeasurements, dispatch_uid='measurements_updated')
```

kacperd:
I have the same issue when deleting user. I have following model:

`class Phone(models.Model):
    user = models.ForeignKey(User, on_delete=models.CASCADE)
    history = HistoricalRecords()

    os = models.CharField(max_length=255, blank=True, null=True)
    os_version = models.CharField(max_length=255, blank=True, null=True)
    app_version = models.CharField(max_length=255, blank=True, null=True)
    created = models.DateTimeField(auto_now_add=True)
`

And I'm getting: 
[...]

PR #676, opened 2020-06-18
Remove db constraint on historical user to prevent IntegrityError 

keithhackbarth:
Add `db_constraint=False` to the history_user field

## Description
There's a known issue (see link below) where if the user_model has `history = HistoricalRecords` tracked on it, then an IntegrityError is raised when trying to delete that user. I believe this is because of the order in which Django deletes essentially causes a circular-referencing error within the database.

This change simply removes the db_constraint from this field. Thus, solving the problem. Multiple other users have raised this as the solution in the linked issue.

I was originally hesitate about use db_constraint=False and was considering making it optional, but noticed that this paradigm is already leveraged multiple times in this codebase.

## Related Issue
https://github.com/jazzband/django-simple-history/issues/391

## How Has This Been Tested?
Added a test to the test suite that covers the bug found in #391. 
Also, it seems like myself (as well as multiple other users) are already running this fork in their own production environments without issue.

n/a

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -67,7 +67,6 @@ class HistoricalRecords(object):
         verbose_name=None,
         bases=(models.Model,),
         user_related_name="+",
-        user_db_constraint=True,
         table_name=None,
         inherit=False,
         excluded_fields=None,
@@ -83,6 +82,7 @@ class HistoricalRecords(object):
         history_user_setter=_history_user_setter,
         related_name=None,
         use_base_model_db=False,
+        user_db_constraint=True,
     ):
         self.user_set_verbose_name = verbose_name
         self.user_related_name = user_related_name
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -67,6 +67,7 @@ class HistoricalRecords(object):
         verbose_name=None,
         bases=(models.Model,),
         user_related_name="+",
+        user_db_constraint=True,
         table_name=None,
         inherit=False,
         excluded_fields=None,
@@ -85,6 +86,7 @@ class HistoricalRecords(object):
     ):
         self.user_set_verbose_name = verbose_name
         self.user_related_name = user_related_name
+        self.user_db_constraint = user_db_constraint
         self.table_name = table_name
         self.inherit = inherit
         self.history_id_field = history_id_field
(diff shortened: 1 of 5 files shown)


rossmechanic:
@keithhackbarth  thanks for this fix! Looks like it failed tests though.

rossmechanic:
Only other comment would to make this something that you can opt in to to maintain backward compatibility (and prevent having to run migrations on a minor version change). Happy to make it the default in version 3.0 when I have a few other backward incompatible changes I want to add. What do you think?

keithhackbarth:
Thanks @rossmechanic 

- Fixed the broken test. Sorry about submitting a broken pull request. This one in particular is a little funky because the core of the issue is foreign key database constraints (MySQL and Postgres). SQLite doesn't enforce this constraint in the same way. The tests always passed on SQLite regardless. To reproduce, I had to use a Postgres test runner.

- Based on your comment re: backward compatibility, made it an optional parameter that defaults to original "True" value

rossmechanic:
@keithhackbarth Thanks for the quick responses here. Going to take a look now. I know - it's been on my list for months to get to adding more DBs to the test suite but haven't had the time as of late.

simple_history/tests/tests/test_models.py
```diff
@@ -370,6 +370,29 @@ def test_specify_history_user(self):
...
+        # Change back to user1
+        document.changed_by = user1
+        document.save()
+
+        # Deleting user1 will cascade delete the document,
+        # but fails when it tries to make the historical
+        # record for the deleted user1.
+        # This test performs differently on Postgres vs. SQLite
+        # because of how the two database handle database constraints
+        user1.delete()
```
rossmechanic, line 394: What's the particular exception you're testing won't be thrown here? When testing that an exception won't be thrown, I generally like to follow this pattern:

```
try:
    the_call_that_will_not_raise_exception()
except ErrorThatWillNotBeRaised as e:
    self.fail(e)
```
```

**Comment by rossmechanic** [comment]
Just had one comment on the test - mind also adding to the change log

**Comment by keithhackbarth** [comment]
@rossmechanic - Good call on updating the test and changelog. Both are done!

File: simple_history/models.py [hunk 2]
```diff
@@ -67,6 +67,7 @@ def __init__(
         verbose_name=None,
         bases=(models.Model,),
         user_related_name="+",
+        user_db_constraint=True,
```
rossmechanic, line 70: Last comment: can you move the argument to the bottom? In v3 we'll look into updating how args are passed, but for now, it's feasible that someone could currently be passing args as positional args and this could break for them by changing the order.

Issue #706, opened 2020-08-26
`as_of` fails on a model with an excluded ManyToManyField


jordonwii:
**Describe the bug**
Calling `as_of` on a model with an excluded `ManyToManyField` will fail if the related field has more than one object. This is because of the `values(...).get()` call here, which behaves differently for M2M fields: https://github.com/jazzband/django-simple-history/blob/dded6890ef619aa8e5458d430026a6520fd3b015/simple_history/models.py#L395.

**To Reproduce**
1. Create a model `MyModel` with an excluded ManyToManyField
2. Create > 1 related objects via the M2M field.
3. Try to call `MyModel.history.as_of(<any-date>)`.

I get:
```
MyModel.MultipleObjectsReturned: get() returned more than one MyModel -- it returned 2! 
```

**Expected behavior**
This should succeed without error.

**Environment (please complete the following information):**
 - OS: Ubuntu
 - Django Simple History Version: master
 - Django Version: 3.1
 - Database Version: sqlite3: 3.22.0 2018-01-22 18:45:57 0c55d179733b46d8d0ba4d88e01a25e10677046ee3da1d5b1581e86726f2alt1

PR #707, opened 2020-08-26
Exclude ManyToManyFields when fetching excluded fields.

jordonwii:
## Description
Calling `as_of` on a model with an excluded `ManyToManyField` will fail if the related field has more than one object. This is because of the `values(...).get()` call here, which behaves differently for M2M fields than for normal scalar fields: https://github.com/jazzband/django-simple-history/blob/master/simple_history/models.py#L395.

This fixes this by simply excluding M2M fields from the reconstructed object instance. AFAICT this is safe and cheap, because users can query their M2M using the model manager if they want values from it. 

An alternate fix could change the code to avoid `.get()`, but I don't think we actually want to prefetch all of the values of an M2M field here. 

## Related Issue
#706 

## How Has This Been Tested?
Test included.

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -384,6 +384,8 @@ class HistoricalRecords(object):
                 field.attname: getattr(self, field.attname) for field in fields.values()
             }
             if self._history_excluded_fields:
+                # We don't add ManyToManyFields to this list because they may cause
+                # the subsequent `.get()` call to fail. See #706 for context.
                 excluded_attnames = [
                     model._meta.get_field(field).attname
                     for field in self._history_excluded_fields
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -14,7 +14,7 @@ from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
-from django.db.models import Q
+from django.db.models import ManyToManyField, Q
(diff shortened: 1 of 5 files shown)


simple_history/models.py
```diff
@@ -387,6 +387,7 @@ def get_instance(self):
                 excluded_attnames = [
                     model._meta.get_field(field).attname
                     for field in self._history_excluded_fields
+                    if not isinstance(model._meta.get_field(field), ManyToManyField)
```
rossmechanic, line 390: @jordonwii this works? Looks like this excludes ManyToManyFields from being excluded? The included fields are in `attrs` above

jordonwii: Right, that's intentional - the goal here is to prevent an M2M field from ending up in the `.values(...)` call on line 395, because then the `.get()` will fail.

rossmechanic: Ohh I see. Forgot about this code. That makes sense. Want to just add a comment here briefly explaining that?

jordonwii: Done. Let me know if it's unclear - wording "exclude a field from the exclusion list we're trying to re-include" is tricky. :)

rossmechanic:
@jordonwii this looks good to me. Just need to fix conflicts

jordonwii:
Great! The conflict was only in CHANGES.rst. It should be resolved now.

PR #709, opened 2020-08-27
Bulk create and update support for alternative managers

### LINKED ISSUE #703: `bulk_update_with_history` to support alternative model managers
**Problem Statement**
I have some models where the default model manager (`objects`) doesn't include all objects (it doesn't include "inactive"/deleted objects). However, `bulk_update_with_history` assumes that the model manager to use is `objects` which means it can never update those objects which are not included.

A way to pass an alternative model manager.

Two options I can think of:

- Pass optional model manager (backwards compatible):
```
def bulk_update_with_history(
    objs, model, fields, batch_size=None, default_user=None, default_change_reason=None, manager=None,
):
    history_manager = get_history_manager_for_model(model)
    with transaction.atomic(savepoint=False):
        model_manager = manager or model.objects
        model_manager.bulk_update(
            objs, fields, batch_size=batch_size,
        )
        history_manager.bulk_history_create(
            objs,
            batch_size=batch_size,
            update=True,
            default_user=default_user,
            default_change_reason=default_change_reason,
        )
```
- Pass manager instead of model (breaking):
```
def bulk_update_with_history(
    objs, model_manager, fields, batch_size=None, default_user=None, default_change_reason=None,
):
    history_manager = get_history_manager_for_model(model_manager.model)
    with transaction.atomic(savepoint=False):
        model_manager.bulk_update(
            objs, fields, batch_size=batch_size,
        )
        history_manager.bulk_history_create(
            objs,
            batch_size=batch_size,
            update=True,
            default_user=default_user,
            default_change_reason=default_change_reason,
        )
```

georgek:
## Description

This adds an optional `manager` parameter to both `bulk_create_with_history` and `bulk_update_with_history`. This is necessary when using models with alternative managers (either because they don't have an `objects` or because `objects` returns a filtered set).

## Related Issue
fixes #703 

## Motivation and Context

Django doesn't require models to have a manager called `objects` and if they do it doesn't necessarily return all objects (so bulk update will not work properly).

## How Has This Been Tested?

Added tests for the functionality and possible exception.

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/utils.py
+++ b/simple_history/utils.py
@@ -5,7 +5,7 @@ from django.db import transaction
 from django.db.models import ManyToManyField
 from django.forms.models import model_to_dict
 
-from simple_history.exceptions import NotHistoricalModelError
+from simple_history.exceptions import AlternativeManagerError, NotHistoricalModelError
 
 
 def update_change_reason(instance, reason):
@@ -78,9 +78,11 @@ def bulk_create_with_history(
         if isinstance(field, ManyToManyField)
     ]
     history_manager = get_history_manager_for_model(model)
+    model_manager = model._default_manager
+
     second_transaction_required = True
     with transaction.atomic(savepoint=False):
-        objs_with_id = model.objects.bulk_create(objs, batch_size=batch_size)
+        objs_with_id = model_manager.bulk_create(objs, batch_size=batch_size)
         if objs_with_id and objs_with_id[0].pk:
             second_transaction_required = False
             history_manager.bulk_history_create(
@@ -100,7 +102,7 @@ def bulk_create_with_history(
                         model_to_dict(obj, exclude=exclude_fields).items(),
                     )
                 )
-                obj_list += model.objects.filter(**attributes)
+                obj_list += model_manager.filter(**attributes)
             history_manager.bulk_history_create(
                 obj_list,
                 batch_size=batch_size,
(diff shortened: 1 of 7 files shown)


simple_history/utils.py
```diff
@@ -134,15 +141,20 @@ def bulk_update_with_history(
...
         record
+    :param manager: Optional model manager to use for the model (default is `objects`)
     """
     if django.VERSION < (2, 2,):
         raise NotImplementedError(
             "bulk_update_with_history is only available on "
             "Django versions 2.2 and later"
         )
     history_manager = get_history_manager_for_model(model)
+    model_manager = manager or model.objects
```
rossmechanic, line 152: Think you can make it the default manager by default rather than objects. That's my bad for not catching that  https://docs.djangoproject.com/en/3.1/topics/db/managers/#default-managers

georgek: Ah yeah, that's even better then. I wonder if it's then not necessary to include the optional parameter to `bulk_create_with_history` because a manager will always be available via the default? The only reason I added it there is in case `objects` doesn't exist.

`bulk_update_with_history` will still require the parameter.

rossmechanic: Yep, that works for me.

georgek: Made the changes and updated the docs.

Issue #732, opened 2020-11-05
Support ignore_conflicts parameter in bulk_create_with_history


shihanng:
**Problem Statement**
Django 2.2 includes `ignore_conflicts` parameter into `bulk_create()`. Would it be possible for `django-simple-history` to allow the same parameter to be passed into `bulk_create_with_history()`?

Documentation: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.bulk_create

**Describe the solution you'd like**
Update `bulk_create_with_history()` with following signature:

```python
def bulk_create_with_history(
    objs,
    model,
    batch_size=None,
    ignore_conflicts=False,
    default_user=None,
    default_change_reason=None,
    default_date=None,
):
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #733, opened 2020-11-05
Add ignore_conflicts to bulk_create_with_history

shihanng:
## Description

If we are to support `ignore_conflicts`, we need to consider that not all item in the list of objects will be bulk created due to conflicts. Then, we also need to make sure that we create history records only for those objects that we created successfully. Unfortunately, we can't rely on the return value of `bulk_create` API because it returns all objects including those it ignores. Therefore, whenever `ignore_conflicts=True` we need to "force" the second transaction to query the created objects.

## Related Issue

Close https://github.com/jazzband/django-simple-history/issues/732

## Types of changes

## Checklist:

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -110,6 +110,7 @@ Authors
 - `Paulo Peres <https://github.com/PauloPeres>`_
 - `Alex Todorov <https://github.com/atodorov>`_
 - David Smith (`smithdc1 <https://github.com/smithdc1>`_)
+- Shi Han Ng (`shihanng <https://github.com/shihanng>`_)
 
 Background
 ==========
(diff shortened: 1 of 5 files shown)


shihanng:
Hi, I am opening this PR to propose a solution to #732. Please take a look and let me know what you think? Thank you.

ThePumpingLemma:
@shihanng would you mind rebasing? Think it looks good otherwise but I'll let @rossmechanic chime in.

shihanng:
@ThePumpingLemma, rebased. Thank you.

ThePumpingLemma (approved):
Thanks!

ThePumpingLemma:
Not sure code climate is ever going to run. /shrug

PR #735, opened 2020-11-06
fix: do not load historic in memory in clean_old_history command

vied12:
uses `query.count()` instead of `len(query)` to count entries to be deleted

## Description

**clean_old_history** command used to count entries to delete with `len()`. This loads in memory all the entries and can lead to crash in case memory limit is reached. Using a sql query to count to entries is faster and safer.

## Types of changes

## Checklist:


ThePumpingLemma (approved):
Good catch!

Issue #758, opened 2020-11-25
Queryset as_of history performance issues


TyrantWave:
**Describe the bug**
The current implementation of a queryset as_of does the checks for changes (And removal of "-" history types) purely in a python set & generator - looping over every single item in the original queryset.

On sufficiently large querysets, this can take a significant amount of time to get a full set of historical models for a queryset at a given point in time.

**To Reproduce**
Have a large (Hundreds of thousands of records or more) queryset, and attempt to work on the entire queryset (For example, generating a set of diffs from a point of time to current over a whole set)

On my data, with ~47,000 items and ~80,000 historical rows, getting a set of changes for one filtered set can take up to 20 minutes.

(I'm aware of the other ticket to add filtering to these querysets and am planning on making a pull request for that in the future)

**Expected behavior**
Generation of the history should be done utilising the database where possible for significant speed gains.

**Environment (please complete the following information):**
 - OS: Mac / Ubuntu 18.04 Docker
 - Django Simple History Version: 2.8.0 ~ latest
 - Django Version: 3.0.3
 - Database Version: PostgreSQL 11

jeking3:
Does the composite indexing I added in #863 do anything to address this?

PR #759, opened 2020-11-25
Queryset history "as_of" speed improvements

TyrantWave:
Now calculates the differences purely in-database, offering a significant speed increase over python.

## Description
Queryset history _as_of_set now uses in-db methods to calculate history at a given time.

## Related Issue
https://github.com/jazzband/django-simple-history/issues/758

## Motivation and Context
Performance increase on large datasets.
On a test set of data with ~46,000 current items and ~80,000 historical entries, time needed to get the full set of changes and filter them reduced from ~15 minutes to ~2 seconds.

## How Has This Been Tested?
Tested against test suite, all passed.

Tested with large datasets, confirmed the same data is returned with & without the change.

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. - No change in functionality, so the same documentation applies.
- [ ] I have added tests to cover my changes. -- No change in functionality, so the same tests apply.

### DIFF
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -92,7 +92,7 @@ class HistoryManager(models.Manager):
         queryset = self.get_queryset().filter(history_date__lte=date)
         latest_pk_attr_historic_ids = (
             queryset.filter(**{pk_attr: OuterRef(pk_attr)})
-            .order_by("-pk")
+            .order_by("-history_date", "-pk")
             .values("pk")[:1]
         )
         latest_historics = queryset.filter(
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -1,4 +1,4 @@
-from django.db import models, connection
+from django.db import connection, models
 from django.db.models import OuterRef, Subquery
 from django.utils import timezone
 
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.db.models import OuterRef, Subquery
 from django.utils import timezone
 
 from simple_history.utils import get_change_reason_from_object
@@ -89,14 +90,17 @@ class HistoryManager(models.Manager):
         model = type(self.model().instance)  # a bit of a hack to get the model
         pk_attr = model._meta.pk.name
         queryset = self.get_queryset().filter(history_date__lte=date)
-        for original_pk in set(queryset.order_by().values_list(pk_attr, flat=True)):
-            changes = queryset.filter(**{pk_attr: original_pk})
-            last_change = changes.latest("history_date")
-            if changes.filter(
-                history_date=last_change.history_date, history_type="-"
-            ).exists():
-                continue
-            yield last_change.instance
+        latest_pk_attr_historic_ids = (
+            queryset.filter(**{pk_attr: OuterRef(pk_attr)})
+            .order_by("-pk")
+            .values("pk")[:1]
+        )
+        latest_historics = queryset.filter(
+            history_id__in=Subquery(latest_pk_attr_historic_ids)
+        )
+        adjusted = latest_historics.exclude(history_type="-").order_by(pk_attr)
+        for historic_item in adjusted:
+            yield historic_item.instance
 
     def bulk_history_create(
         self,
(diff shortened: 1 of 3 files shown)


rossmechanic (changes requested):
Think the ordering is wrong but otherwise looks good. Can't believe we ever did this in code 😢

rossmechanic (commented):
Also looks like it's not supported in MySQL. Can add a conditional or see how we might implement it with mysql...

simple_history/manager.py
```diff
@@ -89,14 +90,17 @@ def _as_of_set(self, date):
...
-            last_change = changes.latest("history_date")
-            if changes.filter(
-                history_date=last_change.history_date, history_type="-"
-            ).exists():
-                continue
-            yield last_change.instance
+        latest_pk_attr_historic_ids = (
+            queryset.filter(**{pk_attr: OuterRef(pk_attr)})
+            .order_by("-pk")
+            .values("pk")[:1]
```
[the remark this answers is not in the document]
TyrantWave: Interestingly I did originally order by history_date, however the tests set all histories to the same date which meant the first ID was chosen, failing them.

I have adjusted to be `order_by("-history_date", "-pk")` to get the most recent via date, largest ID if same date - shouldn't typically be a case that needs to be resolved, but works.

TyrantWave:
> Also looks like it's not supported in MySQL. Can add a conditional or see how we might implement it with mysql...

I'm not overly familiar with this codebase - is there a decent method for switching depending on DB backend?

If so, my own implementation of this ticket uses `DISTINCT ON` for PostgreSQL -> I could add it in to this, which is a better performance gain just for that DB.

TyrantWave:
> Also looks like it's not supported in MySQL. Can add a conditional or see how we might implement it with mysql...

Added checks for MySQL and PostgreSQL specific implementations.

simple_history/manager.py
```diff
@@ -89,14 +90,38 @@ def _as_of_set(self, date):
...
+        # If using MySQL, need to get a list of IDs in memory and then use them for the
+        # second query.
+        # Does mean two loops through the DB to get the full set, but still a speed
+        # improvement.
+        backend = connection.vendor
+        if backend == "mysql":
+            history_ids = {}
+            for item in queryset.order_by("-history_date", "-pk"):
+                if getattr(item, pk_attr) not in history_ids:
+                    history_ids[getattr(item, pk_attr)] = item.pk
```
rossmechanic, line 102: not sure I understand this part?

simple_history/manager.py
```diff
@@ -89,14 +90,38 @@ def _as_of_set(self, date):
...
+            )
+        else:
+            latest_pk_attr_historic_ids = (
+                queryset.filter(**{pk_attr: OuterRef(pk_attr)})
+                .order_by("-history_date", "-pk")
+                .values("pk")[:1]
+            )
+            latest_historics = queryset.filter(
+                history_id__in=Subquery(latest_pk_attr_historic_ids)
+            )
```
rossmechanic, line 121: Why are these two different cases? Postgres can handle subqueries

rossmechanic (commented):
Sorry for delayed response. Two more clarifying questions and then good to go

simple_history/manager.py
TyrantWave: The original method pulled a set of `pk_attr`, then for each one in the set did a filter & latest to get the most recent historical item per `pk_attr`.

This way instead only loops through once - ordering the history set by `-history_date,-pk` (for same date conflicts), and taking the first historical entry found for each `pk_attr`, putting it into the dict to use in the next part.

Hope that explained it well.

simple_history/manager.py
TyrantWave: It can, but the postgres specific one uses `distinct(pk_attr)`, which could potentially have optimizations from the DB engine itself over a subquery - I admittedly don't know enough about Postgres's internals to know if there IS a performance gain for distinct over a subquery to be fair!

Your call on whether to use that or just use a subquery here.

rossmechanic (approved):
lgtm! Thanks @TyrantWave

PR #776, opened 2021-01-06
Diff improvements (performance + `included_fields`)

RealOrangeOne:
## Description

Reduce the number of queries needed when calling `diff_against` (down to 0). 

Also implement `included_fields`, so only certain fields will be compared. In this new implementation of diffing, old fields won't be compared at all, and can be deferred from querying if needed.

## Related Issue

N/A for performance improvement.

`included_fields` was mentioned in https://github.com/jazzband/django-simple-history/pull/576#issuecomment-584761832, and as I was here, I added. Can extract into a separate PR if needed.

## Motivation and Context

`model_to_dict` will do queries for M2M fields against real tables. As the history tables don't use actual relationships, it's safe to use those for the lookups without additional queries. Removing this removes accidental queries on models with M2M fields, and stops loading them into memory unnecessarily.

Additionally, diffing involved getting real instances of the model using `.instance`, which was unnecessary. The diffing can be using the history instances directly, saving a lot of computation. The fields list is pulled off the real model which means `history_*`fields are still excluded correctly. `.instance` will sometimes do queries, hence removing the need for this can additionally improve performance.

There are also additional tests to check that diffing is pure and doesn't touch the DB, to help prevent regressions. 

## How Has This Been Tested?

- [x] Tests have been added to cover changes

- [x] Has been run against live project unittests successfully

## Types of changes

## Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Unreleased
 - Fixed bug where serializer of djangorestframework crashed if used with ``OrderingFilter`` (gh-821)
 - Fixed `make format` so it works by using tox (gh-859)
 - Fixed bug where latest() is not idempotent for identical ``history_date`` records (gh-861)
+- Support ``included_fields`` for ``history.diff_against`` (gh-776)
+- Improve performance of ``history.diff_against`` by reducing number of queries to 0 in most cases (gh-776)
 
 3.0.0 (2021-04-16)
 ------------------
--- a/docs/history_diffing.rst
+++ b/docs/history_diffing.rst
@@ -20,3 +20,5 @@ This may be useful when you want to construct timelines and need to get only the
     delta = new_record.diff_against(old_record)
     for change in delta.changes:
         print("{} changed from {} to {}".format(change.field, change.old, change.new))
+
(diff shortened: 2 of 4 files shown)


RealOrangeOne:
Turns out the current code also causes additional queries for excluded fields, which isn't good. Would be great to get this merged and released soon, as the performance improvements are vast!

sebashwa:
Hey, I created a similar change and then just saw this one.
+1 for integrating ASAP

What do you think about always excluding m2m-fields as they are not supported anyways @RealOrangeOne:

e.g.
```python
def diff_against(self, old_history, excluded_fields=None)
    ---snip---
    for model_field in self.instance._meta.get_fields():
        if model_field.many_to_many:
            excluded_fields.append(model_field.name)
    ---snip---
```

sebashwa (dismissed):

jeking3:
If you could rebase on master and make sure all the new branches are covered by tests, will re-review.

jeking3 (changes requested):
I would leave the KeyError handling in place for invalid fields and add this test to the HistoricalRecordsTest:

```
+        with self.assertRaises(KeyError):
+            delta = new_record.diff_against(old_record, included_fields=["this_field_does_not_exist"])
```
Also rebase on current master, then if we get a clean run we'll be all set.

jeking3 (approved):

PR #791, opened 2021-02-05
Optimise next prev record

RealOrangeOne:
## Description

Improve performance of `next_record` and `prev_record`.

## Motivation and Context

Using `.instance` will run an additional query to get the value for fields which are excluded from history. These properties don't need that, they only need `_meta`, which is accessible from the `.instance_type` without running any queries.

By executing fewer queries, the properties get faster. This can be especially important if the columns contain a large amount of data.

## How Has This Been Tested?

I've manually applied this patch to a project, but also added relevant unit tests (and fixed an existing one which wasn't running).

N/A

## Types of changes

## Checklist:


RealOrangeOne:
Unfortunately it seems this causes issues where history of different models ends up in `prev_record` etc. Will update.

jeking3:
I applied the tests to 3.0.0 (master) and found them to be failing with too many queries, when I applied the changes in the pull request they correctly reduced.  Removing one `Q(...)` does not seem like it would do this however so I'm not sure at the moment where the optimization actually came from, but the tests seem to prove it as without the change it is two queries and with the change it is one.

jeking3 (dismissed):

jeking3 (approved):

PR #807, opened 2021-03-09
Rename Django's dev branch to main.

jezdez:
More information: https://groups.google.com/g/django-developers/c/tctDuKUGosc/
Refs: https://github.com/django/django/pull/14048

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,7 +75,7 @@ jobs:
     - name: Install system dependencies
       run: |
         sudo apt-get -qq update
-        sudo apt-get -y install python-dev python3-dev libmariadbclient-dev default-libmysqlclient-dev build-essential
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -5,19 +5,19 @@ on: [push, pull_request]
 jobs:
   build:
     name: build (Python ${{ matrix.python-version }}, Django ${{ matrix.django-version }})
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
```
jezdez, line 8: This is basically to work around an issue with installing mysqlclient on Ubuntu 20.04 that I couldn't fix:

```
ImportError: libmysqlclient.so.20: cannot open shared object file: No such file or directory
```

ThePumpingLemma (approved):

.github/workflows/test.yml
ThePumpingLemma: Hmm, I can look into this later.

PR #822, opened 2021-04-24
Fix bug where descriptor class missed instance=None case

### LINKED ISSUE #821: djangorestframework 3.12.3 introduced bug where HistoricalModel cannot be used with OrderingFilter
**Describe the bug**
When using a historical model in djangorestframework versions 3.12.3 and newer you cannot use `OrderingFilter` for the `ListAPIView`. If so an exception get raised in the serializer.

**To Reproduce**
1. Create model with history.
2. Create djangorestframework serializer for HistoricalModel.
3. Create a ListAPIView with OrderingFilter.
4. Head over to a link (eg `/api/systems/configs-history/36`) and see the error message.

example:

```py
# urls.py

...
    path('configs-history/<int:config_id>', HistoricalConfigListAPIView.as_view(), name='configs-view'),
...

# views.py

from .serializers import HistoricalConfigSerializer
from rest_framework import filters, generics


class HistoricalConfigListAPIView(generics.ListAPIView):
    serializer_class = HistoricalConfigSerializer
    filter_backends  = [filters.OrderingFilter]

    def get_queryset(self):
        config_obj_id = self.kwargs['config_id']
        config = Config.objects.get(config_id=config_obj_id)
        return config.history.all()
```

**Expected behavior**
No exception when you do this.

**Error logs:**
https://i.imgur.com/J1ga93D.png
https://termbin.com/9fzh

**Reason of the bug (did some debugging for you):**
The internals of `OrderingFilter` call `getattr` for `history_object` on the historical model class. This resolves to calling `__get__` on `HistoricalObjectDescriptor` (there is some metaprogramming checking all attributes of the class) but because it was called on the class not the instance, this method fails because it doesn't account for the case where instance is None (ie a call made on the class instead of the instance). 

https://docs.python.org/3/reference/datamodel.html#object.__get__

>while instance is the instance that the attribute was accessed through, or None when the attribute is accessed through the owner.

This bug got introduced in:

https://github.com/encode/django-rest-framework/pull/7609

 - OS: Debian 10
 - Browser: Firefox
 - Django Simple History Version: 3.0.0
 - Django Version: 2.2
 - Database Version postgresql: 11.11

**Solution:**
Solution should be simple if I understood it correctly, add `if instance is None: return self` to the `__get__` function and djangorestframework continues without errors when it calls for the `history_object` on the model class. I will make a PR for that. 

**Workaround:**
Internals are only called if ordering_fields are not specified. Specify them to get rid of the bug.

hwalinga:
#### Return self if called on the class instead of instance to not error when used with djangorestframework

Some metaprogramming magic in djangorestframework performs a check on getting all the `property` object attributes on a django model. As this loops over all attributes it will also arrive at the `history_object` which is implemented with a descriptor. However, this implementation does not account for being accessed on the class instead of the model. (So `instance` is None.) 

This PR just returns self when that happens. This does not interfere with this metaprogramming done, and makes django-simple-history usable again in combination with djangorestframework. 

Returning self in `__get__` when instance is None is the canonical way of doing this, as explained here:

https://stackoverflow.com/questions/29436716/why-do-you-need-if-instance-is-none-in-get-of-a-descriptor-class

## Related Issue
Fixes #821 

## Motivation and Context
Fixed #821 

## How Has This Been Tested?
I have created the fix and installed it in a project that suffered from this bug and it worked. I don't think we should be pulling in djangorestframework for this?

## Types of changes
Descriptor now accounts for being accessed on the class not the instance. 

## Checklist:

- [ ] My change requires a change to the documentation. *(They don't)*
- [ ] I have updated the documentation accordingly. *(No docs need updating)*
- [x] I have added tests to cover my changes. (Manually confirmed fix)

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -579,6 +579,8 @@ class HistoricalObjectDescriptor:
         self.fields_included = fields_included
 
     def __get__(self, instance, owner):
+        if instance is None:
+            return self
         values = {f.attname: getattr(instance, f.attname) for f in self.fields_included}
         return self.model(**values)
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -49,6 +49,7 @@ Authors
 - Hamish Downer
 - Hanyin Zhang
 - Hernan Esteves (`sevetseh28 <https://github.com/sevetseh28>`_)
+- Hielke Walinga (`hwalinga <https://github.com/hwalinga>`_)
 - Jack Cushman (`jcushman <https://github.com/jcushman>`_)
 - James Muranga (`jamesmura <https://github.com/jamesmura>`_)
 - James Pulec
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changes
 Unreleased
 ----------
 
+ - Fixed bug where serializer of djangorestframework crashed if used with ``OrderingFilter`` (gh-821)
+
 3.0.0 (2021-04-16)
 ------------------
 


hwalinga:
@barm Would you be able to take a look at this to merge it? It is just a very minor two lines fix in a wrongly implemented descriptor class.

hwalinga:
@rossmechanic Would you be able to see if this fix can be merged? It are just two lines solving an annoying issue with a wrongly implemented descriptor class. I saw you had previous commits affecting the surrounding lines.

simple_history/models.py
```diff
@@ -579,6 +579,8 @@ def __init__(self, model, fields_included):
         self.fields_included = fields_included
 
     def __get__(self, instance, owner):
+        if instance is None:
```
rossmechanic, line 582: @hwalinga Why would you want to return the descriptor?

rossmechanic: Oh I see your stackoverflow post. Got it.

rossmechanic (approved):

simple_history/models.py
hwalinga: Yes, it is so that any metaprogramming done on the class knows what kind of attribute this is. (ie a descriptor class) You can also return None, but this way you allow anyone inspecting the class (programmatically) to know what is in here. 

That is also the bug I got in DRF. There was some metaprogramming that failed to continue when it got to this attribute.

PR #832, opened 2021-05-13
On off flag

DanialErfanian:
# Description
Add a flag(settings.py) for disabling history logging by simple_history app. It's maybe useful when you want to save history from specific deployment and ignore history in others.

## Motivation and Context
In our case we need to save history for admin panel to be able track objects modified by our operation team but we don't want history of changes made by API calls.


simple_history/models.py
```diff
@@ -470,12 +470,16 @@ def get_meta_options(self, model):
...
     def post_save(self, instance, created, using=None, **kwargs):
+        if not getattr(settings, "SIMPLE_HISTORY_ENABLED", True):
+            return
         if not created and hasattr(instance, "skip_history_when_saving"):
             return
         if not kwargs.get("raw", False):
             self.create_historical_record(instance, created and "+" or "~", using=using)
 
     def post_delete(self, instance, using=None, **kwargs):
+        if not getattr(settings, "SIMPLE_HISTORY_ENABLED", True):
```
rossmechanic, line 481: How would this handle bulk history update and create?

jeking3:
@DanialErfanian there's an open question that needs answering.

simple_history/models.py
DanialErfanian: Thanks for the comment, Fixed.

DanialErfanian:
@jeking3 
I fixed the problem.

jeking3 (approved):

DanialErfanian:
@jeking3 
Thanks for quick response, Do you have any guess for time of the next release?

PR #729, opened 2020-10-23
Add index to `history_date` field to improve performance of `get_next_record()` and `get_prev_record()` methods

nex2hex:
Based on https://github.com/jazzband/django-simple-history/pull/601

## Description
Calling methods `get_next_record()` and `get_prev_record()`  https://github.com/treyhunner/django-simple-history/blob/master/simple_history/models.py#L391 creates a heavy load on database without indexes (full scan)

## Related Issue
https://github.com/treyhunner/django-simple-history/issues/565

## How Has This Been Tested?
Create new migration after changes and test sql queries time

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ x] I have run the `make format` command to format my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] I have added my name and/or github handle to `AUTHORS.rst`
- [x ] I have added my change to `CHANGES.rst`
- [ ] All new and existing tests passed.

PR #862, opened 2021-08-11
Ensure that latest() is idempotent.

### LINKED ISSUE #861: test_sameinstant is not idempotent
**Describe the bug**

Two records with the same history_date cannot be retrieved idempotently with `latest()`.  The `get_latest_by` meta property uses `history_date` which means if multiple records have the same timestamp, the resulting record returned by the database is not predictable (it would depend on the behavior of the database).  If history_date is not indexed the result may be different than if it is indexed, for example (this was observed in #729).  The meta property should include sorting by history_id.

**To Reproduce**

The unit test `test_sameinstant` tests for this.  When indexing was added in #729, the test started to fail until the meta was fixed.

**Expected behavior**

If two records have the same history_date then use a secondary sort to guarantee the response order that gets limited to one record inside django.

 - Django Simple History Version: 3.0.0
 - Django Version: 3.2.6
 - Database Version: sqlite (whatever the tests use) but applies to all

jeking3:
## Description

Two records with the same history_date cannot be retrieved idempotently with latest(). The get_latest_by meta property uses history_date which means if multiple records have the same timestamp, the resulting record returned by the database is not predictable (it would depend on the behavior of the database). If history_date is not indexed the result may be different than if it is indexed, for example (this was observed in #729). The meta property should include sorting by history_id.

## Related Issue

This closes #861 

## Motivation and Context

When latest() is not idempotent, the behavior can change over time (for example adding an index).

## How Has This Been Tested?

The unit tests exercise this behavior.

## Types of changes

## Checklist:


jeking3:
One job in PR #863 just failed because of this issue.  This code is not in that PR.

sarathak (commented):
code look good.

sarathak (approved):

awais786:
@jeking3 is it possible to put this change behind some flag ? It generates so many migrations in our project.

PR #863, opened 2021-08-11
Add index to `history_date` field to improve performance.

### LINKED ISSUE #565: Add index to `history_date` field
Methods `get_next_record()` and `get_prev_record()` filter table rows by `history_date` field. Adding an index to this field can improve performance of these methods

```
extra_fields = {
    ...
    "history_date": models.DateTimeField(db_index=True)
    ...
}
```

### LINKED ISSUE #758: Queryset as_of history performance issues

jeking3:
## Description

The field `history_date` has never been indexed, yet it is heavily relied on for lookups.  I followed the discussion in #565 and #729.  I split the latest() fix out of #729 and placed it into #862 so that the change to indexing history_date could be done in isolation.  Given the critical importance of indexing on `history_date` I have made it configurable but as an opt-out, not an opt-in; hopefully that is an acceptable compromise.

The as_of queries use a combination of the history_date and model pk, so a composite index on those two fields should also yield gains.  The `SIMPLE_HISTORY_DATE_INDEX` can be set to `composite` to enable this behavior.

## Related Issue

This closes #565.
This closes #729.
This closes #758.

## Motivation and Context

There are _many_ queries that filter on history_date, many using `lte` or `ge` type of queries.  Lack of an index can lead to a table scan, or at least needing to load in every record that meets the other filters.  The database should be doing this work.  With this change the work in #729 is completed and closes #565.  I have a project where I constrain the history_date to be between two timepoints and this eliminates loading all the Historical records just to find them.

## How Has This Been Tested?

The unit tests exercise `history_date` sufficiently, however I added a couple more for the new branches.

## Types of changes

## Checklist:


jeking3:
One job failed due to not having the fix in #862.  Once that is merged and this is rebased, the issue will resolve.

simple_history/models.py
```diff
@@ -467,6 +484,8 @@ def get_meta_options(self, model):
         meta_fields["verbose_name"] = name
         if self.app:
             meta_fields["app_label"] = self.app
+        if self._date_indexing == "composite":
+            meta_fields["index_together"] = (("history_date", model._meta.pk.attname),)
```
tim-schilling, line 488: I believe this change would necessitate a DB change via a migration. If this is enabled, does Django complain about model changes that don't have migrations created for it?

tim-schilling (commented):
Does this approach allow for a developer to change their mind about wanting to have an index? If they use the library and have the index created, will the library determine how to drop that index?

I'd vote to have the index on the field exist by default then leave it up to the developer to add their own migration within their project to remove it if they'd like, avoiding the settings. I think this approach allows the library to sidestep the above scenario, reducing the scope of work. The documentation could even include the migration for dropping the index.

jeking3:
Django manages indexes.  If you change the settings and run "make migrations", indexes on your historical tables will change appropriately.  There is no need for custom migrations.  If you switch from "enabled" to "composite" your django-generated migration will A) drop the single column index; then B) make a composite index.

simple_history/models.py
jeking3: See below, it does not.

jeking3:
@tim-schilling I added a response to your questions, please review.

tim-schilling:
@jeking3 thank you for answering my questions. It looks like I was wrong in my thinking, I appreciate the clarifications. I'll take another look at the PR when I have time later this week. In the meantime, it does look like you committed the merge conflict syntax in `changes.rst`.

jeking3:
I resolved the CHANGES.rst conflict this morning after merging someone else's PR.

jeking3:
Aha, I see, yes there were a couple commits in my branch; I rebased before squashing them, and that crept in.  Fixed that up.

jeking3:
Just rebased on master to fix up the codecov report.

CHANGES.rst
```diff
@@ -4,13 +4,22 @@ Changes
...
-- Fixed ``prev_record`` and ``next_record`` performance when using ``excluded_fields`` (gh-791)
+Upgrade Implications:
+
+- Run `makemigrations` after upgrading to realize the benefit of indexing changes.
+
+Full list of changes:
+
+- Added index on `history_date` column; opt-out with setting `SIMPLE_HISTORY_DATE_INDEX` (gh-565)
+- Added `included_fields` for `history.diff_against` (gh-776)
+- Added Czech translations (gh-885)
```
tim-schilling, line 15: Is this correct?

tim-schilling (changes requested):
I'd suggest making the changes to `changes.rst` outside of the scope of adding indexes on `history_date` in a new PR.

The code otherwise looks good! I will appreciate the work here a lot.

CHANGES.rst
jeking3: Yes it is correct.  I merged a couple PRs that did not have entries in CHANGES.rst so I added them here.  I will clean this up so you can approve it.

jeking3:
@tim-schilling I minimized the changes to CHANGES.rst per your request.

tim-schilling (approved):
Thanks for that last round of changes. I feel more comfortable in reviewing this now.

imhmdb:
@jeking3 when can we expect this to be released? :(
it's been too long since it was merged

thanks for your amazing work btw <3

Issue #715, opened 2020-09-24
Querying history on a related model with as_of


dwasyl:
Hi all,

I've always made good use of the `as_of` function for pulling up historical instances. However, I was recently trying to pull up a series of related objects as of a particular day, where the same thing for related querysets would have been really useful.

The idea would be something like: `Pizza.toppings.as_of(date)` if the model `Pizza` had a related object `Topping` with a foreign key back to `Pizza`.

In the meantime, I ended up querying the history against the model for anything with that foreign key and then having to do individual `.as_of(date)` lookups for each object. Just seemed like there could be a more efficient way of doing this (or perhaps there already is and I missed it).

rossmechanic:
Planning on working on this in the ticket you linked.

dwasyl:
Great, thanks for the followup!

Issue #354, opened 2018-03-24
Querysets as_for


torticolix:
Hi all

I discovered simple-history a few days ago and it fits my needs : have a snapshot of my models along the time
I tried with a simple item model with name and price
I did not find out how to query historical data to get my queryset in the past

For exemple : how to get Items.objects.filter(xxx) or items.objects.all() on Jan 1st 2018 ?

rossmechanic:
`items.history.as_of(datetime.date(2018, 1, 1))` should work.

torticolix:
Yep, but unfortunately it returns a generator, not usable as it to sort, filter or Q , or even use in a cbv ?

rossmechanic:
The code being hit is here on the manager:
```
    def _as_of_set(self, date):
        model = type(self.model().instance)  # a bit of a hack to get the model
        pk_attr = model._meta.pk.name
        queryset = self.get_queryset().filter(history_date__lte=date)
        for original_pk in set(
                queryset.order_by().values_list(pk_attr, flat=True)):
            changes = queryset.filter(**{pk_attr: original_pk})
            last_change = changes.latest('history_date')
            if changes.filter(history_date=last_change.history_date,
                              history_type='-').exists():
                continue
            yield last_change.instance
```
You're welcome to override this and generate the queryset the way you'd like.

rossmechanic:
#229 should also fix this issue. Planning on getting to that PR soon, but feel free to use it as an example of what you can do locally

rossmechanic:
Covered in #397

adibo:
Either I'm not entirely sure what was meant by `as_for`, I reckon the idea was to extend the `history` query capabilities. If so, here is a recipe that could be used:

1. Create a custom `HistoricalRecords` (overwrite descriptor)
2. Create a custom `HistoryDescriptor` (create manager from HistoryQuerySet)
3. Extend `HistoryManager` (define customized `HistoryQuerySet`, add own query methods)
4. Register your django model using your defined custom classes

Steps 1-3: create e.g. `history.py` in your project structure with the `django-simple-history` customizations:
```
# django-simple-history customizations
from simple_history.models import default_get_user, HistoricalRecords
from simple_history.manager import HistoryDescriptor, HistoryManager
import simple_history.utils as utils

class HistoryQuerySet(models.QuerySet):
    """Customized django-simple-history QuerySet"""
    # extend your QuerySet to your liking
    pass


[...]

Issue #118, opened 2014-09-19
Is there a recommended way to version a *model graph*


estebistec:
I _think_ this is what #17 is getting at, but I thought it was worth asking to be clear.

When models have FK relations, you can put history on both models, but the FK references on history instances will still always point to the primary, non-history objects.

So, e.g.,
- I have model `A` and model `B` with an FK to `A`
- For an instance of `A` with several related instances of `B`
- Where the instances of `B` are being changed

What is the best way to have multiple versions of the `A` instance that is related to the correct historical instances of `B`?

I see that the history manager has `as_of`, which takes a date-time, but that is of course potentially non-exact if two versions could ever land on the same measurable instant.

Is there a better or more obvious way I'm not seeing? Is there a way to link historical versions of one model to historical versions of another to get _model graph_ snapshots?

matheusjardimb:
I'm not sure if I'm asking the same thing as the stated in the question, but this is my situation:

```
class Product(Model):
    history = HistoricalRecords()
    name = CharField(max_length=200, verbose_name=_('name'))

class Order(Model):
    history = HistoricalRecords()
    company = ForeignKey('Company', verbose_name=_('company'), on_delete=PROTECT)
    customer = ForeignKey('Customer', verbose_name=_('customer'), on_delete=PROTECT)
    purchased_at = DateTimeField(verbose_name=_('purchased at'), null=True, blank=True)

class OrderItem(DateModel):
    history = HistoricalRecords()
    order = ForeignKey('Order', verbose_name=_('order'), on_delete=PROTECT)
    product = ForeignKey('Product', verbose_name=_('product'), on_delete=SET_NULL)
    quantity = PositiveIntegerField(verbose_name=_('quantity'), default=1)
```

The store admin may change the product name over time (or even delete it) but I need to show to the customer the information exactly as when he bought it.

By using the code below, I'm able to get only the historic version of Order, but OrderItem and Product data retrieved are not the historical ones.
```
order = Order.objects.first()
as_bought = order.history.as_of(order.order_created_at)
```

Let me know if this post is not related, so that I can create a new issue.

Thanks!

adibo:
I came across a similar use case and agree that it would be great to have a version of `as_of` which produces a historical snapshot of the model instance itself (which it does) with historical versions of all related models. I don't think it is possible today (or trivial to implement).
I guess, for now, you should be able to live with something like:

```
order_items = OrderItem.objects.filter(order_id=as_bought.id) 
[...]

PR #864, opened 2021-08-11
Add queryset filtering on HistoryManager with `as_of` support.

### LINKED ISSUE #397: Allow filtering of as_of queryset

### LINKED ISSUE #803: What is the recommended way to get the most recent HistoricalModel instance?

jeking3:
## Description

The discussion in #397 is about using filtering and querysets with point-in-time queries.  The current `as_of` implementation returns a generator so it can only be used to make a single query and obtain the result.  It does convert those results to the genuine model.  The ideal solution would allow someone to express the following:

```python
    queryset = MyModel.history.filter(as_of=<datetime>, ...)
```

That's what this pull request provides.  One of the challenges with `as_of` filtering is handling deletions; the logic is not trivial and requires a subquery.  When querying this way you get a queryset that returns Historical models, unless you tell it you want genuine instances:

```python
    queryset = MyModel.history.filter(as_of=<datetime>, ...).as_original()
```

Calling `as_original` on the queryset will cause it to return genuine instances instead of historical records.  You can still query directly on `history_date` if you want to be able to retrieve deletion records.  You cannot query on as_of with any qualifiers like `__lte`.

I had originally made the as_of query a method on the HistoryManager called `as_of_filter`.  One benefit to this approach is that if the original model had a field named `as_of` there would not be a conflict.  However it seems more natural to allow an as_of filter expression, so that's what I'm submitting as a solution.

Also note there is a small test change to deal with #862 before it gets merged.

## Related Issues

#118
#229
#354
This closes #397
#715
#758
This closes #803

## Motivation and Context

## How Has This Been Tested?

Tests were added, all tests pass.  Additionally, I am integrating the changes into a prototype that was working based on work in #354.

## Types of changes

## Checklist:

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -56,6 +56,7 @@ Authors
 - Jesse Shapiro
 - Jihoon Baek (`jihoon796 <https://github.com/jihoon796>`_)
 - Jim Gomez
+- Jim King (`jeking3 <https://github.com/jeking3>`_)
 - Joao Junior (`joaojunior <https://github.com/joaojunior>`_)
 - Joao Pedro Francese
 - `jofusa <https://github.com/jofusa>`_
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changes
 Unreleased
 ----------
 
- - Fixed bug where serializer of djangorestframework crashed if used with ``OrderingFilter`` (gh-821)
+- Fixed bug where serializer of djangorestframework crashed if used with ``OrderingFilter`` (gh-821)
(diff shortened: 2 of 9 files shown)


jeking3:
Sorry about the churn, I decided after the initial PR to make it so one could not call `as_original()` on the HistoricalQuerySet once it fetched.

jeking3:
The latest force push allows `as_of=None` to be passed into filter() without incident (it is effective ignored, whereas before it would cause an error).  I caught this while re-integrating with my project.

jeking3:
While this passes tests, I'm running into problems with _as_of_filter keeping the settings while integrating, so working on that.

jeking3:
Will re-submit when I work out the kinks.

PR #865, opened 2021-08-12
Add a queryset based `as_of` and get history or instances.

### LINKED ISSUE #397: Allow filtering of as_of queryset

### LINKED ISSUE #803: What is the recommended way to get the most recent HistoricalModel instance?

jeking3:
## Description

The discussion in #397 is about using filtering and querysets with point-in-time queries.  The current `as_of` implementation returns a generator so it can only be used to make a single query and obtain the result.  It does convert those results to the genuine model.  The documentation states that it must return an iterable, so it does not have to be a generator; it could be a queryset.  Additionally, people may want to query historical records directly without the translation using similar `as_of` logic.

In this pull request, the following changes were made:

1. HistoryManager.as_of returns a queryset instead of a generator.  Both are iterable, no tests had to change.  I don't consider this a breaking change.
2. HistoryManager uses a new HistoricalQuerySet, so if you use `MyModel.history.filter(a=b)` you get a HistoricalQuerySet.
3. HistoricalQuerySet has an as_of method, which applies the same as_of logic (without dropping deletions) as the HistoryManager.
4. HistoricalQuerySet has an `as_instances` method, which returns a queryset that returns instances instead of historical records.  When you enable it, it drops any deletion records from the queryset.
5. When a HistoricalQuerySet is in instance mode, filtering on `pk` will be applied to the original type's private key instead of the history_id field.

This allows you to use as_of as a filter and then continue to filter, or do anything else a queryset can do, and then you can decide if you want historical records or instances as a result.

Example (from unit tests):

``` python
    def test_historical_queryset(self):
        t0 = datetime.now()
        document1 = RankedDocument.objects.create(rank=42)
        document2 = RankedDocument.objects.create(rank=84)
        t1 = datetime.now()
        document2.rank = 51
        document2.save()

        # nothing exists at t0 (this call goes to the legacy HistoryManager method)
        uut = RankedDocument.history.as_of(t0)
        self.assertEqual(list(uut), [])

        # Leverage the new filter-based as_of directly; this is
        # returning historical records and not instances
        uut = RankedDocument.history.all().as_of(t1)
        self.assertEqual(len(list(uut)), 2)
        self.assertEqual([item.history_type for item in uut], ["+", "+"])
        # remember historical records come back in reverse chronological order:
        self.assertEqual([item.rank for item in uut], [84, 42])

        # It's a regular query so you can filter it further
        uut2 = uut.filter(rank__gte=50)
        self.assertEqual(len(list(uut2)), 1)
        self.assertEqual(uut2[0].rank, 84)

        # `as_instances` lets us convert a historical queryset to return instances
        # instead of historical records, and it's a queryset so you can keep filtering
        uut = uut.as_instances().order_by("rank")
        self.assertEqual(len(list(uut)), 2)
        self.assertEqual(list(uut), [document1, document2])

        # prove that HistoryManager.as_of is the same as the newer filter methods
        # (internally, it uses them, so it should be!)
        self.assertEqual(
            list(RankedDocument.history.as_of(t1)),
            list(RankedDocument.history.all().as_of(t1).as_instances()),
        )

        document1.delete()
        t2 = datetime.now()

        # at t2 we see one instance, the other has been deleted
        uut = RankedDocument.history.as_of(t2)
        self.assertEqual(list(uut), [document2])

        # if we use queryset based as_of we can see the mod and the deletion
        uut = RankedDocument.history.all().as_of(t2)
        self.assertEqual(len(list(uut)), 2)
        self.assertEqual(list(uut)[0].history_type, "-")
        self.assertEqual(list(uut)[1].history_type, "~")

        # however if we then use as_instances() on it, we do not see deleted instances
        uut = uut.as_instances()
        self.assertEqual(list(uut), [document2])

        # as_instances is idempotent
        uut = uut.as_instances()
        self.assertEqual(list(uut), [document2])
```

Also note there is a small test change to deal with #862 before it gets merged.

## Related Issues

#118
#229
#354
This closes #397
#715
#758
This closes #803

## Motivation and Context

## How Has This Been Tested?

Tests were added, all tests pass.  Additionally, I am integrating the changes into a prototype that was working based on work in #354.

## Types of changes

## Checklist:

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -56,6 +56,7 @@ Authors
 - Jesse Shapiro
 - Jihoon Baek (`jihoon796 <https://github.com/jihoon796>`_)
 - Jim Gomez
+- Jim King (`jeking3 <https://github.com/jeking3>`_)
 - Joao Junior (`joaojunior <https://github.com/joaojunior>`_)
 - Joao Pedro Francese
 - `jofusa <https://github.com/jofusa>`_
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changes
 Unreleased
 ----------
 
- - Fixed bug where serializer of djangorestframework crashed if used with ``OrderingFilter`` (gh-821)
+- Fixed bug where serializer of djangorestframework crashed if used with ``OrderingFilter`` (gh-821)
+- Added queryset-based filtering with ``as_of`` (gh-397)
+- Added option to return historical records or instances with querysets (gh-397)
+- Added ``pk`` translation when queryset is in instance mode (gh-397)
 
 3.0.0 (2021-04-16)
 ------------------
--- a/docs/querying_history.rst
+++ b/docs/querying_history.rst
@@ -95,19 +95,68 @@ This will change the ``poll`` instance to have the data from the
 ``HistoricalPoll`` table indicating that a new change has been made.
 
 
-as_of
------
+as_of, as_instances
+-------------------
 
-This method will return an instance of the model as it would have existed at
-the provided date and time.
+The HistoryManager allows you to query a point in time for the latest historical
+records.  The following examples come from the manager unit tests.
 
 .. code-block:: pycon
 
-    >>> from datetime import datetime
-    >>> poll.history.as_of(datetime(2010, 10, 25, 18, 4, 0))
-    <Poll: Poll object as of 2010-10-25 18:03:29.855689>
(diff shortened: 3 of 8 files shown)


jeking3:
It looks like sqlite3 and postgres are working, but mysql and mariadb cannot handle the changes, which caused all the failures.  Digging into it.

jeking3:
Looks like I will need to do some more work on this in order to support mysql and mariadb, so I will open a new PR for it when ready.

PR #866, opened 2021-08-13
Extend `as_of` to return a queryset instead of a list of instances (both are iterable).

### LINKED ISSUE #397: Allow filtering of as_of queryset

### LINKED ISSUE #803: What is the recommended way to get the most recent HistoricalModel instance?

jeking3:
## Description

The discussion in #397 is about using filtering and querysets with point-in-time queries.  The current `as_of` implementation returns a instance generator so it can only be used to make a single query and obtain the result. The documentation states that it must return an iterable, so it does not have to be a generator.  As a queryset, no tests needed to change to handle it.

HistoricalQuerySet is now returned from queries on history.  This is used internally by as_of queries to prune results to the latest of each primary key, and to convert to instances. 

The resulting queryset from as_of can be further filtered; any use of `pk` will be converted to the original model's primary key if the queryset is returning instances.

Example (from unit tests):

``` python
    def test_as_of(self):
        """Demonstrates how as_of works now that it returns a QuerySet."""
        t0 = datetime.now()
        document1 = RankedDocument.objects.create(rank=42)
        document2 = RankedDocument.objects.create(rank=84)
        t1 = datetime.now()
        document2.rank = 51
        document2.save()
        document1.delete()
        t2 = datetime.now()

        # nothing exists at t0
        queryset = RankedDocument.history.as_of(t0)
        self.assertEqual(queryset.count(), 0)

        # at t1, two records exist
        queryset = RankedDocument.history.as_of(t1)
        self.assertEqual(queryset.count(), 2)
        self.assertEqual(queryset.filter(rank__gte=75).count(), 1)
        ids = set([item["id"] for item in queryset.values("id")])
        self.assertEqual(ids, {document1.id, document2.id})

        # at t2 we have one record left
        queryset = RankedDocument.history.as_of(t2)
        self.assertEqual(queryset.count(), 1)
        self.assertEqual(queryset.filter(rank__gte=75).count(), 0)

```

## Related Issues

#118
#229
#354
This closes #397
#715
#758
This closes #803

## Motivation and Context

## How Has This Been Tested?

Tests were added, all tests pass.  Existing tests did not have to change.  Additionally, my company has been using this in production for over a month.

## Types of changes

## Checklist:

### DIFF
--- a/docs/querying_history.rst
+++ b/docs/querying_history.rst
@@ -98,16 +98,44 @@ This will change the ``poll`` instance to have the data from the
 as_of
 -----
 
-This method will return an instance of the model as it would have existed at
-the provided date and time.
+The HistoryManager allows you to query a point in time for the latest historical
+records or instances.  When called on an instance's history manager, the ``as_of``
+method will return the instance from the specified point in time, if the instance
+existed at that time, or raise DoesNotExist.  When called on a model's history
+manager, the ``as_of`` method will return instances from a specific date and time
+that you specify, returning a queryset that you can use to further filter the result.
 
 .. code-block:: pycon
 
-    >>> from datetime import datetime
-    >>> poll.history.as_of(datetime(2010, 10, 25, 18, 4, 0))
-    <Poll: Poll object as of 2010-10-25 18:03:29.855689>
-    >>> poll.history.as_of(datetime(2010, 10, 25, 18, 5, 0))
-    <Poll: Poll object as of 2010-10-25 18:04:13.814128>
+    >>> t0 = datetime.now()
+    >>> document1 = RankedDocument.objects.create(rank=42)
+    >>> document2 = RankedDocument.objects.create(rank=84)
+    >>> t1 = datetime.now()
+
+    >>> RankedDocument.history.as_of(t1)
+    <HistoricalQuerySet [
+        <RankedDocument: RankedDocument object (1)>,
+        <RankedDocument: RankedDocument object (2)>
+    ]>
+
+    >>> RankedDocument.history.as_of(t1).filter(rank__lte=50)
+    <HistoricalQuerySet [
+        <RankedDocument: RankedDocument object (1)>
+    ]>
+
+``as_of`` is a convenience: the following two queries are identical.
+
+.. code-block:: pycon
+
+    RankedDocument.history.as_of(t1)
+    RankedDocument.history.filter(history_date__lte=t1).latest_of_each().as_instances()
+
+If you filter by `pk` the behavior depends on whether the queryset is
+returning instances or historical records.  When the queryset is returning
+instances, `pk` is mapped to the original model's primary key field.
+When the queryset is returning historical records, `pk` refers to the
+`history_id` primary key.
+
 
 most_recent
 -----------
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -1,6 +1,6 @@
 from django.conf import settings
 from django.db import connection, models
-from django.db.models import OuterRef, Subquery
+from django.db.models import OuterRef, QuerySet, Subquery
 from django.utils import timezone
 
 from simple_history.utils import (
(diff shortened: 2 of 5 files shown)


jeking3:
I have found that, while this does work for basic usage, you can get into a situation where you retrieve a historical record (or an instance) and then try to access a related table through dot notation, and if that was since deleted, it will fail.

jeking3:
Waiting for #863 to merge before I rebase this one.  Could someone please review it?  Thanks.

jeking3:
I believe I found an issue at https://github.com/jazzband/django-simple-history/pull/866/files#diff-4b2bcf1fafd0c47640f253c43a1cb48ac5b9a244a013d704192ec3a33e1dd146R174 whereby if `instance=False` then what we want is a `queryset.filter(history_date__lte=as_of)` instead of calling the HistoricalQuerySet._as_of which does additional instance based filtering to remove deletions.

jeking3:
The next push resolves the issue I found (and the incorrect test assertions), and makes the code more understandable.

jeking3:
Hmm, I think perhaps as_of should always perform a latest_for_each() instead of what I pushed.  If someone wants multiple historical records for the same object they can filter on `history_date__lte` directly.  Fixing that up.  The logic was actually correct before in that respect. :|

jeking3:
Forgot to modify the documentation appropriately for the last push.

jeking3:
That last push was to cover an idempotence branch in HistoricalQuerySet.as_instances() - looks like it was successful.

jeking3:
I changed as_instances() to use the original table's ordering, however that ordering can refer to other non-historical tables by relation, which will not have linkage, and fail.  Removing the ordering; as_of instances will come back in reverse chronological order.

jeking3:
I have been using this code in production for a few weeks, it's been stable and effective at allowing for queries into history with additional filters.

jeking3:
Rebased now that the history_date PR was merged.

jeking3:
Well shoot, one of the recent test assertions on the number of queries for previous and next caught something; looking into it.

jeking3:
I split the test runner changes into their own PR.

jeking3:
@inexcode give it a try.

jeking3:
@jazzband/django-simple-history is this project no longer maintained?  Nobody's looking at pull requests.

simple_history/manager.py
```diff
@@ -9,13 +9,114 @@
...
+        """
+        Ensures results in the queryset are the latest historical record for each
+        primary key.  Deletions are not removed.
+
+        Returns a queryset.
+        """
+        # If using MySQL, need to get a list of IDs in memory and then use them for the
+        # second query.
+        # Does mean two loops through the DB to get the full set, but still a speed
+        # improvement.
```
tim-schilling, line 63: I didn't realize this wasn't new code. You can ignore the following if you'd like.


A speed improvement in what case(s)? If we don't have a repeatable test case to prove that, this comment isn't helpful.

Ideally, this should use the same code for all vendors. If there's a bug or performance problem, allowing the developer a hook to change how the query is built may be better.

Personally, I'd vote for the subquery approach.

simple_history/manager.py
```diff
@@ -9,13 +9,114 @@
...
+            )
+            latest_historics = self.filter(history_id__in=latest_pk_attr_historic_ids)
+        else:
+            latest_pk_attr_historic_ids = (
+                self.filter(**{self._pk_attr: OuterRef(self._pk_attr)})
+                .order_by("-history_date", "-pk")
+                .values("pk")[:1]
+            )
+            latest_historics = self.filter(
+                history_id__in=Subquery(latest_pk_attr_historic_ids)
```
tim-schilling, line 85: I didn't realize this wasn't new code. You can ignore the following if you'd like.

Since you have an outer ref to `self._pk_attr` whih is doing association, you can have the following:

```
self.filter(Exists(latest_pk_attr_historic_ids))
```

simple_history/manager.py
```diff
@@ -9,13 +9,114 @@
...
+
 class HistoryDescriptor:
     def __init__(self, model):
         self.model = model
 
     def __get__(self, instance, owner):
         if instance is None:
-            return HistoryManager(self.model)
+            return HistoryManager.from_queryset(HistoricalQuerySet)(self.model)
         return HistoryManager(self.model, instance)
```
tim-schilling, line 120: How come we don't want to use the `from_queryset` when instantiating the manager with an instance?

simple_history/manager.py
```diff
@@ -66,16 +167,37 @@ def most_recent(self):
         return self.instance.__class__(**values)
 
     def as_of(self, date):
-        """Get a snapshot as of a specific date.
+        """
+        Get a snapshot as of a specific date.
+
+        The resulting queryset will locate the most recent historical record
+        before the specified date for each primary key, generating instances.
```
tim-schilling, line 174: This function returns either a queryset or an instance (I'm not a fan of that, it should only do one or the other preferably, but that's how it was built originally). This variation should continue to be mentioned in the docstring somewhere. I first interpreted this such that the function would always return a queryset.

simple_history/manager.py
```diff
@@ -66,16 +167,37 @@ def most_recent(self):
...
+        If the most recent historical record is a deletion, that instance is
+        dropped from the result.
```
tim-schilling, line 176: I don't understand what this actually means. Can you clarify here and maybe together we can come up with alternative language?

docs/querying_history.rst
```diff
@@ -98,16 +98,42 @@ This will change the ``poll`` instance to have the data from the
 as_of
 -----
 
-This method will return an instance of the model as it would have existed at
-the provided date and time.
+The HistoryManager allows you to query a point in time for the latest historical
+records or instances.  The ``as_of`` method will return instances from a specific
+date and time that you specify, and return a queryset that you can use to further
+filter the result.
```
tim-schilling, line 104: Doesn't `as_of` either return a single instance or a queryset? If so, I think this should be reworded a little to make that clear.

> The ``as_of`` method will return an instance or a queryset from a specified date and time.

Ideally we'd call out when it will return a instance vs queryset too.

tim-schilling (changes requested):
The code changes here are minimal. My concerns are around the language of the documentation and comments. Please be aware that two of my comments can be ignored entirely as it's related to old code that's simply being moved, but I wanted to raise it incase you agree.

I could be misunderstanding some of the comments or code, so if you think that's the case please let me know and we can work through it.

tim-schilling:
@jeking3 We should do a release before this gets merged in. This change necessitates a major version change as it's changing the API.

simple_history/manager.py
jeking3: If you are querying for instances (using the as_of query), and the latest historical record (based on your timepoint in the query) is a deletion record, that instance is not included in the result.  This behavior is identical to what was there before.

If you are querying for historical records, and the latest historical record (based on your timepoint in the query) is a deletion record, you will get it.

docs/querying_history.rst
jeking3: If you explore the history of the related issues, you will see most people wanted to leave as_of alone and use something else to represent a proper timepoint query.  I agree with you that as_of returning an instance or a queryset is confusing.  Previously it returned either an instance or an array of instances.

jeking3:
My intention was that the documented interface for as_of declares the return be an iterable.  It used to be a list of instances.  It is now a queryset; both of which are iterable.  From a documentation perspective they are not different.

simple_history/manager.py
jeking3: Most of the subquery code was simply moved into HistoricalQuerySet so it could be more selectively applied; and was not otherwise changed.

simple_history/manager.py
jeking3: Most of the subquery code was simply moved into HistoricalQuerySet so it could be more selectively applied; and was not otherwise changed.

simple_history/manager.py
jeking3: Good question, I believe it would be useless to get a historical queryset when instantiating from an instance because HistoricalQuerySet exists to facilitate being able to query history or instances; if you are already at an instance then you do not need to query history.  That said, the better answer here is "I don't know" as my implementation and tests haven't needed it.  as_of queries in all my use cases are always on an model and not on an instance.

simple_history/manager.py
jeking3: Yes, it used to return one instance or a list of instances.  I wouldn't mind separating the concern here and actually using a breaking change to fix it up...

tim-schilling: I'm getting up to speed on the library. I think what you have here is fine considering it's similar to existing functionality.

simple_history/manager.py
tim-schilling: Alright, thank you for the explanation.

jeking3:
I'm really starting to see this problem better now, in that "as of" is a global time point for the query _and anything that follows it_ including chasing relations.  I actually have a solution working against this now in my fork and in my product, for example when two tables are historic and one refers to the other, I added a HistoricForeignKey that is used in place of ForeignKey that is smart enough to honor the referrer's timepoint and can chase the relationship properly (the code is not in this PR; something I have to work on getting into this project).  This behavior relies on the code in this PR adding more information to objects coming out of instanceize(), namely `_history` is a property with the historic record that made the instance, and `_history`.`_as_of` is the timepoint of the query that generated the instance from which the relation is being chased.

In all what this PR plus that additional work provides is foreign keys that work naturally between two tables that capture history.  So not only would we get queryset results that we can filter, but also the ability to chase relationships correctly from the `as_of` timepoint in the queryset.

jeking3:
Understood.  It's probably best to address your concerns here and then open a separate pull request for issue-880.

jeking3:
CI success blocked on #938

tim-schilling (dismissed):
I'm good with this for as much as I'm familiar with the codebase.

jeking3:
@tim-schilling sorry to ask, I had to rebase and it resets the review; could you re-approve?

LeMinaw (approved):

PR #885, opened 2021-09-18
Added: Czech translate

rdurica:
Added czech translate
## Description

Created translates for czech language

## Screenshots (if appropriate):
Screenshot from 2021-09-18 23-06-57

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have run the `make format` command to format my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] I have added my name and/or github handle to `AUTHORS.rst`
- [ ] I have added my change to `CHANGES.rst`
- [x] All new and existing tests passed.

PR #889, opened 2021-09-29
add missing changelog for gh-885

jeking3:
@tim-schilling if you could approve this as well I would appreciate it.

#885 did not note this in the changelog.


tim-schilling (approved):
Thanks for clearing this up!

Issue #919, opened 2021-11-23
Allow Overriding admin templates


benf710:
**Problem Statement**
In base Django you are able to override admin templates to implement additional customizations. This doesn't work for the templates that are being used for django-simple-history.

**Describe the solution you'd like**
django-simple-history should allow you to override the base admin templates if needed in the same way that Django does.

**The Solution**
The way that Django implements this is by passing multiple template names to the templating engine. Django will try them in order until one of them resolves and will use that one.

**Additional context**
[Django Docs: Admin | Overriding admin templates.](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#admin-overriding-templates)
[Django Admin Source lines](https://github.com/django/django/blob/8b020f2e64f1cbf2b06205a389a13af6623f90ce/django/contrib/admin/options.py#L1960)

PR #920, opened 2021-11-23
Passing multiple template names to the templating engine. Django will…

benf710:
## Description 
This passes the Django default admin template override paths to the views in order so that a user can override the django admin templates used in this project.

## Related Issue
This addresses #919 

## How Has This Been Tested?
Tested with a Django app to override the `object_history.html` template per the Django documentation

## Types of changes
This change does not introduce any breaking changes in any version of Django >= 1.10. It allows the end user implementing this library to implement custom templates in their own project per Django documentation: https://docs.djangoproject.com/en/dev/ref/contrib/admin/#overriding-admin-templates

## Checklist:

- [ x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/simple_history/admin.py
+++ b/simple_history/admin.py
@@ -20,8 +20,6 @@ SIMPLE_HISTORY_EDIT = getattr(settings, "SIMPLE_HISTORY_EDIT", False)
 
 
 class SimpleHistoryAdmin(admin.ModelAdmin):
-    object_history_template = "simple_history/object_history.html"
-    object_history_form_template = "simple_history/object_history_form.html"
 
     def get_urls(self):
         """Returns the additional urls used by the Reversion admin."""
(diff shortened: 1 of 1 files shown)


jeking3:
@benf710 maybe rebasing this on master will help?

simple_history/admin.py
```diff
@@ -20,9 +20,6 @@
 
 
 class SimpleHistoryAdmin(admin.ModelAdmin):
-    object_history_template = "simple_history/object_history.html"
-    object_history_form_template = "simple_history/object_history_form.html"
```
jeking3, line 24: Based on CI, something depends on this (could just be tests - didn't look).

jeking3 (changes requested):
CI broke with the changes.

PR #923, opened 2021-11-25
Fix diff_against on model with non-editable fields

dracos:
## Description

#776 improved the performance of `diff_against` by setting `fields` directly to the fields of the instance. But `model_to_dict` ignores non-editable fields, which leads to an error when it loops over `fields` and can't fetch the non-editable fields.

I have added a test that fails without my change to ignore non-editable fields.

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/tests/models.py
+++ b/simple_history/tests/models.py
@@ -25,6 +25,14 @@ class Poll(models.Model):
         return reverse("poll-detail", kwargs={"pk": self.pk})
 
 
+class PollWithNonEditableField(models.Model):
+    question = models.CharField(max_length=200)
+    pub_date = models.DateTimeField("date published")
+    modified = models.DateTimeField(auto_now=True, editable=False)
+
+    history = HistoricalRecords()
+
+
 class PollWithUniqueQuestion(models.Model):
     question = models.CharField(max_length=200, unique=True)
     pub_date = models.DateTimeField("date published")
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -78,6 +78,7 @@ from ..models import (
     Place,
     Poll,
     PollInfo,
+    PollWithNonEditableField,
     PollWithExcludedFieldsWithDefaults,
     PollWithExcludedFKField,
     PollWithExcludeFields,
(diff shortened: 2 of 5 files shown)


thijskramer:
Nice! I happened to stumble on this problem as well, so I'm happy that you've fixed it!

jeking3 (dismissed):
@dracos this just needs to be rebased on master.

dracos:
Hopefully done :)

simple_history/tests/models.py
```diff
@@ -25,6 +25,14 @@ def get_absolute_url(self):
         return reverse("poll-detail", kwargs={"pk": self.pk})
 
 
+class PollWithNonEditableField(models.Model):
+    question = models.CharField(max_length=200)
+    pub_date = models.DateTimeField("date published")
+    created = models.DateTimeField(auto_now_add=True, editable=False)
```
jeking3, line 31: By making this field `auto_now_add`, it is only touched at create time.  If you change it to `auto_now` it will better test the behavior.

simple_history/tests/tests/test_models.py
```diff
@@ -697,6 +698,18 @@ def test_history_diff_with_included_fields(self):
         self.assertEqual(delta.changed_fields, ["question"])
         self.assertEqual(len(delta.changes), 1)
 
+    def test_history_diff_with_non_editable_field(self):
+        p = PollWithNonEditableField.objects.create(
+            question="what's up?", pub_date=today
+        )
+        p.question = "what's up, man?"
+        p.save()
```
jeking3, line 706: Given the field `PollWithNonEditableField.created` is designated as `auto_now_add`, save() will not change it, so this test does not really exercise the exclusionary effect of `editable=False`.

jeking3 (changes requested):
I didn't notice this before, sorry, but since `auto_now_add` is being used, `created` is not actually changing in the test, so it isn't really testing the change.

simple_history/tests/tests/test_models.py
dracos: I'm happy to change this, and have now done so, but I would note the test was already testing the behaviour, because it doesn't matter if the save changes the field, the comparison of all fields is still performed.

jeking3 (approved):

PR #938, opened 2022-01-02
test against django 4.0 proper

jeking3:
We were testing in ci against django 4.0rc1 and not django 4.0.  There was new code in djmain that caused one of our tests to fail.

This closes #937

PR #941, opened 2022-01-12
Add HistoricForeignKey for chasing relationships through an as_of timepoint.

### LINKED ISSUE #17: Add support for as_of_related on HistoryManager
As implemented in this fork: https://bitbucket.org/joaofrancese/django-simple-history/src/0d1630812d2e/simple_history/manager.py?at=default

### LINKED ISSUE #118: Is there a recommended way to version a *model graph*

### LINKED ISSUE #880: Chasing ForeignKeys across a historic timepoint
**Problem Statement**
I found an interesting way to allow historic instances to load their foreign keys by value, assuming the foreign key's model is also being handled by django-simple-history.  I wanted to share, and see if it sparks any further ideas from people who know the code better.

**Related Issues**

I found a couple issues that are related.  Essentially what I did in this issue as a thought exercise allows you to use foreign keys of as_of instances, as long as both tables use simple history.

#118
#17

If table A and B are both historic, and A has a foreign key relationship to B (let's call this field `foo`), simple_history will drop the foreign key but maintain the _id field.  If one queries the history on table A and gets an instance, if the corresponding record in table B has already been deleted then you end up with a `DoesNotExist` error.

What I would like to see is that if you delete the record from B, then try to call A.history.earliest().instance.foo, it would find the corresponding record from B before it was deleted and return an instance of it.

**How I solved this locally**
First, I augmented `get_instance` to inject the historical record into the resulting instance as `_history`:

```
$ git diff before..after
diff --git a/simple_history/models.py b/simple_history/models.py
index de3762b..dc5763b 100644
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -32,6 +32,9 @@ except ImportError:
 
 registered_models = {}
 
+# when converting a historical record to an instance, this attribute is added
+# to the instance so that code can reverse the instance to its historical record
+SIMPLE_HISTORY_REVERSE_ATTR_NAME = "_history"
 
 def _default_get_user(request, **kwargs):
     try:
@@ -393,7 +397,11 @@ class HistoricalRecords:
                     pass
                 else:
                     attrs.update(values)
-            return model(**attrs)
+            result = model(**attrs)
+            # this is the only way external code could know an instance is historical
+            setattr(result, SIMPLE_HISTORY_REVERSE_ATTR_NAME, self)
+            return result
 
         def get_next_record(self):
             """
diff --git a/simple_history/tests/tests/test_models.py b/simple_history/tests/tests/test_models.py
index bb8abef..5516e3b 100644
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -16,7 +16,7 @@ from django.urls import reverse
 
 from simple_history import register
 from simple_history.exceptions import RelatedNameConflictError
-from simple_history.models import HistoricalRecords, ModelChange
+from simple_history.models import HistoricalRecords, ModelChange, SIMPLE_HISTORY_REVERSE_ATTR_NAME
 from simple_history.signals import pre_create_historical_record
 from simple_history.tests.custom_user.models import CustomUser
 from simple_history.tests.tests.utils import (
@@ -359,6 +359,13 @@ class HistoricalRecordsTest(TestCase):
             },
         )
 
+    def test_reverse_historical(self):
+        """Tests how we can go from instance to historical record."""
+        document = Document.objects.create()
+        historic = document.history.all()[0]
+        instance = historic.instance
+        self.assertEqual(getattr(instance, SIMPLE_HISTORY_REVERSE_ATTR_NAME).history_date, historic.history_date)
+
     def test_specify_history_user(self):
         user1 = User.objects.create_user("user1", "1@example.com")
         user2 = User.objects.create_user("user2", "1@example.com")
```

Once that was done, I tried to find a way to augment django without monkey patching it, but I wasn't able to as of yet.  I was trying to replace the Meta._base_manager so that it would return history if a historic instance, or objects if not.  This would not be sufficient however since you need to run an `as_of` query.  So what I did instead is this rather hacky solution:

```
from django.db.models.fields.related_descriptors import ForwardManyToOneDescriptor
from django.db.models.fields.related_descriptors import ReverseOneToOneDescriptor
from simple_history.models import SIMPLE_HISTORY_REVERSE_ATTR_NAME

original_rood_get_queryset = getattr(ReverseOneToOneDescriptor, "get_queryset")
original_fmod_get_queryset = getattr(ForwardManyToOneDescriptor, "get_queryset")


def historic_aware_rood_get_queryset(rood: ReverseOneToOneDescriptor, **hints):
    instance = hints.get("instance")
    if instance:
        history = getattr(instance, SIMPLE_HISTORY_REVERSE_ATTR_NAME, None)
        if history:
            return rood.related.related_model.history.as_of(history.history_date)
    return original_rood_get_queryset(rood, **hints)


setattr(ReverseOneToOneDescriptor, "get_queryset", historic_aware_rood_get_queryset)


def historic_aware_fmod_get_queryset(fmod: ForwardManyToOneDescriptor, **hints):
    instance = hints.get("instance")
    if instance:
        history = getattr(instance, SIMPLE_HISTORY_REVERSE_ATTR_NAME, None)
        if history:
            return fmod.field.remote_field.model.history.as_of(history.history_date)
    return original_fmod_get_queryset(fmod, **hints)


setattr(ForwardManyToOneDescriptor, "get_queryset", historic_aware_fmod_get_queryset)
```

I put that code into a file and force it to load during serilalizer `__init__` in my project.  Now when the code in django's relation processing asks for a queryset, it will get back a queryset that has been `as_of` filtered using the `history_date` that was in the historical record behind the instance.  This also has a benefit of allowing you to know if an instance is derived from history or not, which might be useful as well.  I noticed this does not work for chained lookups however, for example if you have a `get_object_or_404` that uses a filter like  `b__c__id` it doesn't seem to see "b" in this case.  I'm sure with another patch that could be made to work as well.

Any thoughts on how to improve this would be quite welcome.  Thanks.

jeking3:
## Description

Built on top of #866 (which is currently the base for this pull request), these changes
enable ForeignKey lookups across historic instances from an as_of timepoint.  Consider
the following:

1. Model "Organization" is historic.
2. Model "Member" is historic with a foreign key to Organization.
3. At t1, Member "m1" is created in an organization.
4. At t2, Member "m2" is created in an organization.
5. At t3, Member "m1" is removed from an organization.

By changing Member's ForeignKey pointing to Organization into a HistoricForeignKey, you can retrieve the organization at a time point and then chase the members at that time point more naturally:

```
    members_at_t1 = organization.as_of(t1).members.all()
    # members_at_t1 will contain m1
    members_at_t2 = organization.as_of(t2).members.all()
    # members_at_t2 will contain m1 and m2
    members_at_t3 = organization.as_of(t3).members.all()
    # members_at_t3 will contain m2
```

To make this query before, one would have to query Members with as_of, which retrieves a list, then filter it.  The combination of #866 and these changes completes foreign key support with timepoint access.  It works through forward and reverse relationships.

This work does not specifically address M2M relationships, however this is a good foundation for completing that work.  Making the M2M table historic and changing the foreign keys to HistoricForeignKey might just do it...

## Related Issues

This closes #880 
This closes #118 
This closes #17 

## Motivation and Context

Querying foreign key relationships from history was rather unnatural.  This makes it quite natural.

## How Has This Been Tested?

1. Unit tests included.
2. Already in use in a production product.

## Types of changes

## Checklist:

- [X] I have run the `pre-commit run` command to format and lint.


jeking3:
This is blocked by #938

jeking3:
@tim-schilling this one is ready for review

jeking3:
It seems nobody wants to touch this with a 10 foot pole... but it does work. :)

marco-silva0000:
@jeking3 for what it's worth, I've skimmed trough the code and it looks reasonable, the feature itself is really needed, great work.

jeking3:
@jazzband/django-simple-history would appreciate a review and (hopefully) approval so this can be merged.

jeking3:
@jazzband/django-simple-history would appreciate a review and (hopefully) approval so this can be merged.

utapyngo:
A stupid question. How to make this relation work historically using a lookup? 
Does this PR covers this use case?
I want to find teams whose lead first name was 'Alice' at `ts`.


```
class Employee(AbstractUser):
    ...
simple_history.register(Employee)

class Team(models.Model):
    # Do I have to use HistoricForeignKey here?
    lead = HistoricForeignKey(Employee, models.SET_NULL)  
    ...
simple_history.register(Team)

lead = create_employee(first_name='Alice')
team = create_team(lead=lead)
ts = timezone.now()
lead.first_name = 'Bob'
lead.save()
Team.history.as_of(ts).filter(lead__first_name='Alice').count()
```

The above code return 0. I expect it to return 1.

jeking3:
HistoricForeignKey makes it easier to chase the foreign key internally but I don't think it is going to assist you with the relation filtering as you have it written.  You could do this instead, using HistoricForeignKey:

```
# get all the teams for lead with first name Alice at timepoint ts
# assumes both `Lead` and `Team` are historic and HistoricForeignKey is used on the relationship
# assumes no related_name is used, hence the `_set` naming convention for the reverse relationship

Lead.history.as_of(ts).filter(first_name="Alice").team_set.all()
```

Both forward and reverse foreign keys (in this PR) will chase the as_of timepoint across the foreign key property accessors.

jeking3:
I had to rebase due to conflicts.

utapyngo (approved):

LuxsSoft:
Hi there, 

This is great, i was trying out this feature if it works for my needs. Whats confusing at first is that both tables have to be historic. All i need is basically to track all the members of an organization at a points in time. So the organization is historic but the members never change themselfs and are thus not historic semantically. The tests also never seem to aim to proof the case of changing member attributes. 

In my case its not members but a set of static through table instances that have to be tracked at a point in time. 
Would it be possible to adopt this to avoid having another historic table for a model that is actually not historic?

PR #944, opened 2022-01-24
add ci job for testing doc build

jeking3:
This adds a CI build step to build the docs so we can look for regressions related to sphinx updates.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,6 +86,11 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r requirements/tox.txt
 
+      - name: Check documentation build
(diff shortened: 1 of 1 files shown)


jezdez:
Hey @jeking3 in theory that is useful, but luckily we don't have to build this on our own but can use Read The Docs' own functionality to achieve per-pull request builds: https://docs.readthedocs.io/en/stable/pull-requests.html I've just enabled that feature in the RTD admin UI. This should work the next time a PR is created/updated.

jeking3:
It did, so I will close this.  Our readme might need to be updated to indicate "make docs" is no longer needed except to test out changes.  We're still getting dependabot updates on sphinx, for example, so this was to ensure that did not rot.

PR #956, opened 2022-02-09
Use pre-commit on tox lint.

hramezani:


jeking3 (approved):
Tried it locally, works.  Stronger lint than before.

Issue #960, opened 2022-03-02
Time for a new release?


1ace:
It's been a while since 3.0.0 and [a lot has been done](https://github.com/jazzband/django-simple-history/compare/3.0.0...master). Is it time for 3.1.0/4.0.0?

In my case, I need https://github.com/jazzband/django-simple-history/pull/866, and I'm finding myself installing `git+https://github.com/jazzband/django-simple-history#master` to be able to use it.

brendanbrett:
I am also eagerly awaiting a new release, specifically for #866! I hope the great work done by @jeking3 can be reviewed and incorporated soon.

1ace:
Pinging @barm who did the last release, and @rossmechanic who did pretty much all the releases before that :slightly_smiling_face:

jeking3:
Thanks for the kind words.  I would really like to see #941 get in as well, which builds on top of #866.  I don't mind attempting to drive a release but I have been waiting for an approval on that one for two months.

lazarust:
Getting #941 and #932 merged in and released would be really great for me! They both have features that I'm really needing.

utapyngo:
Can we release them separately? 
#941 is already in.
#932 has conflicts and broken tests, and the author does not seem to have time to resolve them.

lazarust:
That would be fine with me! I just don't want #932 to be forgotten.

utapyngo:
Pinging @barm who did the last release, and @rossmechanic who did pretty much all the releases before that.

barm:
Hey all, sorry for the radio silence. I should be able to help cut a release this week.

lazarust:
@barm Any update on when you think you'd be able to make that release?

jeking3:
Review and approve, if you could: https://github.com/jazzband/django-simple-history/pull/969

jeking3:
https://github.com/jazzband/django-simple-history/releases/tag/3.1.0

As far as I know, apart from dropping Django 2.2, 3.1, Python 3.6 - this is compatible with 3.0.0.  There was a change around the `as_of` call that now returns a filtered QuerySet instead of a list of results.  The documentation for `as_of` explicitly stated that the return value was an Iterable.  This has not changed.  Both list[] and QuerySet are iterable.  Therefore I tagged this as a pre-release for 3.1.0 and we'll see what happens.  The alternative is to make it a 4.0.0 release.

utapyngo:
When will there be a release to PyPI?

[...]

PR #981, opened 2022-04-29
[fix] add fallback_version to scm settings

areski:
## Description
setuptools_scm [https://pypi.org/project/setuptools-scm/] is causing issues if you are trying to install this package directly from a `zip` or `tar.gz`.
That is expected as there is no metadata in the archive file, and the following error will raise:
```
      LookupError: setuptools-scm was unable to detect version for /tmp/pip-req-build-plm5aze3.
```

This is probably not affecting a lot of people but there is situation where it's not possible to use git to install the package, or wait for the latest version to be on pypi.

# Alternative solutions
* Use env `SETUPTOOLS_SCM_FALLBACK_VERSION`
* Use git archives: https://pypi.org/project/setuptools-scm-git-archive/

## Related Issue
ref: https://github.com/jazzband/django-simple-history/issues/960#issuecomment-1098147525
ref: https://github.com/pypa/setuptools_scm/issues/409

## Proposed Solution
A quick workaround is using `fallback_version` see attached PR,
there is certainly more elegant solution, hopefully one that could still use the latest tag.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ with open("README.rst") as readme, open("CHANGES.rst") as changes:
             "local_scheme": "node-and-date",
             "relative_to": __file__,
             "root": ".",
+            "fallback_version": "0.0.0",
         },
         setup_requires=["setuptools_scm"],
         description="Store model history and view/revert changes from admin site.",


jeking3:
As a workaround you can use the following requirements.txt syntax to pull a hash directly from GitHub:

django-simple-history @ git+https://github.com/jazzband/django-simple-history@<GIT HASH, BRANCH, OR TAG>#egg=django-simple-history-3.1.1+<SOMETHING>

jeking3 (approved):

areski:
Thanks for the merge and the hint, unfortunately I cannot use Git in my , only `tar.gz` or `zip`.
Thanks anyway :)

PR #984, opened 2022-05-07
[issue-983] querying history, convert to instance, chase foreign key on history_date

### LINKED ISSUE #983: Select from Model.history, then use .instance, then chase a foreign key is broken
**Describe the bug**

In 3.1 HistoricForeignKey was introduced which is intended to work with Model.history.as_of(timepoint) and this works.

In 3.1.1, calling `.instance` on a historic record lookup does not populate `_as_of` which is necessary for HistoricForeignKey to work.  You can query history with as_of(timepoint) and things work properly, but querying Model.history directly then converting to an instance cannot chase the foreign keys properly.

My workaround in another project:

```
            if isinstance(value, Value.history.model):
                # work around https://github.com/jazzband/django-simple-history/issues/983
                # queries beginning with as_of properly inject, but direct selections
                # converted to instances do not... :|  I missed something!!
                setattr(value, "_as_of", value.modified_at)
                # then value.instance properly works
```

jeking3:
## Description

This fixes an issue whereby:

1. User queries history directly.
2. User converts Historical<> to instance with instance property.
3. Original query (from 3.1) was not with history.as_of so it did not have _as_of property
4. instance failed

The fix:

When _history._as_of is not present use the history.history_date field instead
when chasing foreign keys.

This closes #983 

In the end, if you did not query history with as_of, and you end up with a historical instance, if you call .instance on it, the HistoricalForeignKey chasers will use a fallback of the history_date of the record that made the instance when chasing relations! This was an easy fix and an easy win, and makes django-simple-history more powerful.

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -673,7 +673,7 @@ class HistoricForwardManyToOneDescriptor(ForwardManyToOneDescriptor):
                 None,
             )
             if history and histmgr:
-                return histmgr.as_of(history._as_of)
+                return histmgr.as_of(getattr(history, "_as_of", history.history_date))
         return super().get_queryset(**hints)
 
 
@@ -708,7 +708,9 @@ class HistoricReverseManyToOneDescriptor(ReverseManyToOneDescriptor):
                         None,
                     )
                     if history and histmgr:
-                        queryset = histmgr.as_of(history._as_of)
+                        queryset = histmgr.as_of(
+                            getattr(history, "_as_of", history.history_date)
+                        )
                     else:
                         queryset = super().get_queryset()
                     return self._apply_rel_filters(queryset)
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -2012,3 +2012,16 @@ class HistoricForeignKeyTest(TestCase):
             to_historic(ot1), TestOrganizationWithHistory.history.model
         )
         self.assertIsNone(to_historic(org))
+
+        # test querying directly from the history table and converting
+        # to an instance, it should chase the foreign key properly
+        # in this case if _as_of is not present we use the history_date
+        # https://github.com/jazzband/django-simple-history/issues/983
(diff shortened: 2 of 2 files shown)


tim-schilling (approved):
This makes sense to me. Good catch!

LuxsSoft:
``This seem not fully resolved.

I had the mentioned error before version 3.2.0, now i can get a  record with something like historicalOrgInstance.history.latest().instance and also historicalOrgInstance.history.latest().prev_record.instance. However getting the instance that way always returns an empty set of historical foreign key models. 

It only works by getting the instance through an as_of() query, then the  reverse relationships models show up through the related name eg. 
historicalOrgInstance.historical_participants.all()

LuxsSoft:
More specifically here is what i am trying to do

t1: org with participant 1

org = HistoricalOrg(name="original")
part1 = HistoricalParticipant.objects.create(name="p1", org=org)
before_mod = timezone.now()

t2: org with participant 2

org.name="modified"
org.save()
org.participants.all().delete()
part2 = HistoricalParticipant.objects.create(name="p1", org=org)

now what works as expected:

org.as_of(before_mod).participants.all() # returns part1
org.as_of(before_mod).name # returns "original"
org.as_of(timezone.now().participants.all() # returns part2
org.as_of(timezone.now()).name # returns "modified"


what is weird when i want to work with the helpers:

org.as_of(bevore_mod).participants.all() # returns part1 as expected as before
org.history..first().instance.name # returns "modified" as expected
org.history..first().instance.participants.all() returns part1 BUT part2 is expected
org.history..first().prev_record.instance.name # returns "original" as expected
org.history..first().prev_record.instance.participants.all() # returns empty HistoricalQuerySet [] BUT part1 is expected

tim-schilling:
@LuxsSoft it would be best to open up a new issue and write a draft PR with the tests that show the problem.

LuxsSoft:
> @LuxsSoft it would be best to open up a new issue and write a draft PR with the tests that show the problem.

ok i tried to make the case

Issue #337, opened 2018-01-04
invalid literal for int() with base 10: 'a string' when using hashid fields.


spyoungtech:
I just ran into this issue using v 1.9.0 -- Just like #182 

I'm using a `HashidAutoField` from [django-hashid-field](https://github.com/nshafer/django-hashid-field)  with my models. This seems to still confuse simple-history in the same way.

For example... a profile model
```python
class UserProfile(models.Model):
    id = HashidAutoField(alphabet=string.ascii_uppercase+string.digits, primary_key=True)
    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name='profile')
    # other fields
    history = HistoricalRecords()
```

Any ideas where I can start tracking down the cause? Looking at the PR that resolved #182 -- a cause or solution was not immediately apparent.

rossmechanic:
Hey @spyoungtech did you ever figure this out?

spyoungtech:
Oh boy, I wish I could remember the complete context around this or the complete steps to reproduce. I am a doofus for not including the entire stack trace.

If memory serves, [models.py L418](https://github.com/treyhunner/django-simple-history/blob/9e95fd8093356986c2eff36247a275fce0c1c3d9/simple_history/models.py#L418) may have been part of the issue.

The HashidAutoField is happy to accept strings, and while under the hood, IntegerField was OK for representation in the DB, caused some issues in migration because the field would return as a string on the original model.


However, I just tried reproducing on latest versions of both packages and either the steps I'm taking are different or the problem may have been solved. Not sure, but will look into it more this weekend.

legshort:
@spyoungtech This is my quick solution, perhaps not for best one cuz I'm pretty sure there is a decode method from `hashidfield` library that converts from str hash code to internal int value but I haven't seen any code working like that at a glance.

```python
class HashIdFieldAdminMixin:
    def _decode_id(self, model, object_id):
        obj = model.objects.all_with_deleted().get(id=object_id)
        decoded_id = str(obj.id.id)

        return decoded_id

    def history_view(self, request, object_id, extra_context=None):
        decoded_id = self._decode_id(self.model, object_id)

        return super().history_view(request, decoded_id, extra_context=extra_context)

    def history_form_view(self, request, object_id, version_id):
        decoded_id = self._decode_id(self.model, object_id)

        return super().history_form_view(request, decoded_id, version_id)
```

legshort:
[...]

PR #1009, opened 2022-07-05
support django-hashid: fix ValueError when using with django-hashid-field

areski:
## Description
A ValueError is raised on the admin when I update an instance, which is due to use the hashid string value instead of the id value.
The goal is to work nicely with https://github.com/nshafer/django-hashid-field

## Related Issue
Ref: https://github.com/jazzband/django-simple-history/issues/337

**To Reproduce**
1. Create the following:
> models.py
class MyModel(models.Model):
    id = BigHashidAutoField(primary_key=True, prefix="ctl_")
    ...

> admin.py
Using `HashIdFieldAdminMixin` referenced at https://github.com/jazzband/django-simple-history/issues/337

@admin.register(MyModel)
class MyModelAdmin(HashIdFieldAdminMixin, SimpleHistoryAdmin):
    ...

2. Edit in django-admin an instance of MyModel
4. See error: `ValueError: Field 'id' expected a number but got 'ctl_182R302JMK4NE'.`

## Types of changes

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -570,7 +570,9 @@ class HistoricalRecords:
 
         attrs = {}
         for field in self.fields_included(instance):
-            if field.attname == instance._meta.pk.attname and hasattr(instance, "id_hashid"):
+            if field.attname == instance._meta.pk.attname and hasattr(
+                instance, "id_hashid"
+            ):
                 hashid = getattr(instance, "id_hashid")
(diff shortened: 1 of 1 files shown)


areski:
This is not needed if you use:
`HASHID_FIELD_ENABLE_HASHID_OBJECT = True`

Closing this PR!

Issue #1014, opened 2022-07-21
Make unneeded count operation optional in `clean-duplicate-history`.


bigtimecriminal:
**Problem Statement**
Performance of the `clean-duplicate-history` function could be optimized.

**Describe the solution you'd like**
I want to suggest a simple improvement. There's nested `count` operation the result of which is only using in a log statement at verbosity level 2: https://github.com/jazzband/django-simple-history/blob/master/simple_history/management/commands/clean_duplicate_history.py#L69

I'd like to run `count` only if we actually log the result. Otherwise we should use `exists`.

**Describe alternatives you've considered**
None. I think this is a pretty simple change

PR #1015, opened 2022-07-21
Use exists query instead of count in clean-duplicate-history command.

bigtimecriminal:
## Description
Instead of using a nested `count` query in `clean-duplicate-history`, use `exists` unless we're at `verbose >= 2`

## Related Issue
https://github.com/jazzband/django-simple-history/issues/1014

## Motivation and Context
Removes costly query that only exists to log something out on a log that might not be called.

## How Has This Been Tested?
Ran all tests. Command is already being tested by test suite

## Types of changes

## Checklist:


rossmechanic (dismissed):
lgtm

ddabble (approved):
Looks like this was simply forgotten about, so I'll merge now :)

Issue #704, opened 2020-08-21
raise PermissionDenied in history_view


rodrigondec:
# I'm willing to open PR

related to #631 ?

**Describe the bug**
In the history view it's verified if it has `self.has_change_permission`

https://github.com/jazzband/django-simple-history/blob/7b1538908bca61ab2c0574c2d44ff5c485faa978/simple_history/admin.py#L73

Why does it need change permission to view the history?

> I'm not talking about the `history_form_view`!

I'm changing the method `has_change_permission` to disable edit's.

I'm receiving a 403 Forbidden because of this!

**To Reproduce**
1. Overwrite the method has_change_permission in the AdminModel

```python
    def has_change_permission(self, request, obj=None):
        """
        Desabilita alteração da entidade
        """
        return False
```
2. Go to admin
3. Go to entry of the model
4. Access `history`

**Expected behavior**
I was expecting to view my history!

**Screenshots**
![image](https://user-images.githubusercontent.com/3301060/90909365-053d3980-e3ac-11ea-8ccd-cbbb8c79385a.png)

![image](https://user-images.githubusercontent.com/3301060/90909379-0bcbb100-e3ac-11ea-9ea0-96974374d058.png)

![image](https://user-images.githubusercontent.com/3301060/90909411-15edaf80-e3ac-11ea-998b-2903efc060e1.png)

## Without the overwrite on has_change_permission
![image](https://user-images.githubusercontent.com/3301060/90909522-3d447c80-e3ac-11ea-8859-09bf834b699d.png)

vied12:
Hello,

For now I'm doing like this : 

```python
def has_change_permission(self, request, obj=None):
        # special rule for history, otherwise it doesn't show up
        if resolve(request.path_info).url_name == 'core_wescaleactivity_history':
            return True
        return False
```

rossmechanic:
Yep this is related to #631 and should be addressed in the next version of DSH

erikvw:
see #1017

PR #631, opened 2020-02-16
Admin revert permissions2

erikvw:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have run the `make format` command to format my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] I have added my name and/or github handle to `AUTHORS.rst`
- [ ] I have added my change to `CHANGES.rst`
- [ ] All new and existing tests passed.

Issue #475, opened 2018-11-02
Utilize has_view_permission to let users with view permissions see the history


dgerritsen:
**Problem Statement**
Currently, a user with view permissions can see the model instance itself, but not the history.

**Describe the solution you'd like**
A user with view, but not change permissions, to be able to view the history.

**Describe alternatives you've considered**
-

**Additional context**
-

erikvw:
What I am proposing / working on in PR #529 might address this.

erikvw:
I have a PR I am submitting now that adds a settings attribute ``SIMPLE_HISTORY_ENFORCE_HISTORY_MODEL_PERMISSIONS``. When set set to ``True``, permissions assigned
to history models are assigned and evaluated explicitly and do not implicitly use the permissions from the concrete model.

This adds flexibility to allow/disallow access to history models by model.

ddabble:
Linking the PR mentioned in the comment above, as it seems to resolve this issue. Please undo if that's not the case!

PR #1017, opened 2022-07-26
evaluate history model permissions explicitly

erikvw:
Add a feature to evaluate history model permissions explicitly when settings.SIMPLE_HISTORY_ENFORCE_HISTORY_MODEL_PERMISSIONS is set to True.

## Description
With the example of the ``Poll`` model, by default granting view and change permissions to the ``Poll`` model
implicitly grants view and change permissions to the ``Poll`` history model. When enabled, this feature expects explicit permissions to be assigned to a user wanting to access a history model. This allows for more granular permission management of the history models. 

The new settings attribute ``SIMPLE_HISTORY_ENFORCE_HISTORY_MODEL_PERMISSIONS`` toggles the feature on. By default the attribute is False and there is no change in behaviour when False.

The feature is built into the ``admin.py``.  The methods ``has_view_permissions`` and ``has_change_permissions`` are no longer shared between the model and the history model. The methods ``history_view`` and the ``history_form_view`` instead access the new methods ``has_view_history_permissions`` and ``has_change_history_permissions`` where permissions for the history model are evaluated instead of the permissions for the model. If the feature is not enabled, both of the new methods just return  ``has_view_permissions`` or ``has_change_permissions``, respectively.

Regardless of the permissions a user has to the model, if the user has only view permissions to the history model, the revert button is hidden (Even if ``SIMPLE_HISTORY_REVERT_DISABLED=False``).

If the user has no permissions to the model but does have, say, view permissions to the history model, the ``app_index`` and ``changelist`` are not accessible, as access to these are controlled by the model's permissions only. I think this is as it should be. If this is how an app manager wants to assign user permissions; that is, to only the history models, then the  app can provide a direct link to the history model instead of navigating through the app_index->changelist->history.

It makes more sense to use this feature with ``SIMPLE_HISTORY_REVERT_DISABLED = False``.
## Related Issue

 #631, #704, #529, #475 

## Motivation and Context
Our application has an auditor role. We have been disabling the REVERT button using the settings attribute, but it recently seemed to stop working as expected. I believe this is because  the ``view`` permission had been somewhat muddled with the ``change`` permission of the model. More importantly though, I want to be able to allow access to some history records and not others based on a user role / group.

## How Has This Been Tested?
- Yes, tests are included and coverage shows the changes are touched. 
- My env is set up similarly to the one used (sort, black, etc).
- As far as I have tested, there is no change to default behavior.

None
## Types of changes

## Checklist:

- [X] I have run the `pre-commit run` command to format and lint.


electis (dismissed):

docs/admin.rst
```diff
@@ -72,12 +72,86 @@ admin class
...
+
+The user below has view and change permissions to the ``Poll`` model and the ``Poll``
+history model in admin.
+
+.. code-block:: python
+
+    user.user_permissions.clear()
+    user.user_permissions.add(
+        Permission.objects.get(codename="view_planet"),
+        Permission.objects.get(codename="change_planet"),
```
merwok, line 108: planet or poll?

erikvw: poll. thanks

akhayyat:
The current implicit permissions based on concrete model change permission is confusing. This change would be very useful. Any reason it's not merged yet?

simple_history/admin.py
```diff
@@ -231,6 +233,54 @@ def content_type_model_cls(self):
...
+    def has_view_permission(self, request, obj=None):
+        return super().has_view_permission(request, obj)
+
+    def has_change_permission(self, request, obj=None):
+        return super().has_change_permission(request, obj)
+
+    def has_view_or_change_permission(self, request, obj=None):
+        return self.has_view_permission(request, obj) or self.has_change_permission(
+            request, obj
+        )
```
ddabble, line 255: I'm curious as to why you added these methods when they provide exactly the same implementation as the super class. Is it to communicate that this can be a good starting point for future changes, should the need ever arise? In that case, it might be a good idea to add some comments 🙂

(This doesn't affect my approval in any way, though)

ddabble (approved):
Thank you, this is absolutely a necessary change, in my opinion!

Assuming the changes I added are still in line with what you had in mind, @erikvw, I'll merge in a couple weeks' time if no remarks have been made - if it's not already been merged by you in the meantime 🙂

Issue #1054, opened 2022-10-31
Document auto_now not updating inside of bulk_update_with_history


twolfson:
**Problem Statement**
When using `bulk_update_with_history`, a `DateTimeField` using `auto_now` doesn't automatically update like it does with `Model.save()`

This is technically a deeper Django issue, however this is good to document in a repo which stands on such bulk update conventions as well as well as using `Model._default_manager`.

**Describe the solution you'd like**
Update "Common Issues" documentation to spread awareness of issue (took us a year to catch) and how to mitigate it

**Describe alternatives you've considered**
I already have a working solution, this is strictly documenting/sharing for others

**Additional context**

PR #1055, opened 2022-10-31
Documented auto_now usage with bulk_update_with_history

twolfson:
## Description
When using `bulk_update_with_history`, a `DateTimeField` using `auto_now` doesn't automatically update like it does with `Model.save()`

This PR documents the issue under "Common Issues" and offers a convention-based solution

## Related Issue

https://github.com/jazzband/django-simple-history/issues/1054

## Motivation and Context

The motivation is to spread awareness of the issue and guide people towards a solution. It seems like a rather quiet bug that can really hurt data accuracy.

## How Has This Been Tested?

The code is excerpts from our code

## Types of changes

## Checklist:

### DIFF
--- a/docs/common_issues.rst
+++ b/docs/common_issues.rst
@@ -303,5 +303,8 @@ As a convention, you can define a custom manager to update these timestamps auto
         pass
 
     class Poll(models.Model):
+        created_at = models.DateTimeField(auto_now_add=True)
+        updated_at = models.DateTimeField(auto_now=True)
+
         # This must be the first Manager declared in this class, otherwise it won't be detected as the `default_manager`
         objects = UpdatedAtFriendlyManager()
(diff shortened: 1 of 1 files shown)


rossmechanic:
@twolfson what's the point? Don't you already have those auto_now dates on the history objects? I've traditionally ignored updated/created columns from my base model in my historical model

twolfson:
I think there's 2 ways to interpret your response so I'll respond to both:

- Why do you need an `updated_at` column on your normal model? (this is the fix I'm proposing to document)
    - This is a great question!
    - It's not used in application code so much, but it's very useful to avoid looking at extra pages or doing complex joins when looking into issues
- Why does the historical model need an `updated_at` column?
    - It doesn't but this issue about that. It's about the first one -- when using `bulk_update_with_history`, we want to automatically update the `auto_now` column on the normal model
    - Arguably this isn't `django-simple-history` responsibility, but that's why it's only being proposed as documentation, not as a code-based solution

Issue #533, opened 2019-03-03
Foreign Key value in __str__ results in error.


dojutsu-user:
**Describe the bug**
I have two models like this:
```python3
class Book(models.Model):
    title = models.CharField(max_length=50)
    # other fields

class BookCode(models.Model):
    book_id = models.CharField(max_length=10)
    associated_book = models.ForeignKey(Book, on_delete=models.CASCADE)
    history = HistoricalRecords()
    # other fields

    def __str__(self):
        return f'[{self.book_id.upper}] {self.associated_book.title}'
```

When I try to get the history of model `BookCode`, this happens
```
>>> BookCode.history.all()
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/pikachu/Desktop/Practice/library/venv/lib/python3.6/site-packages/django/db/models/query.py", line 247, in __repr__
    return '<%s %r>' % (self.__class__.__name__, data)
  File "/home/pikachu/Desktop/Practice/library/venv/lib/python3.6/site-packages/django/db/models/base.py", line 503, in __repr__
    return '<%s: %s>' % (self.__class__.__name__, self)
  File "/home/pikachu/Desktop/Practice/library/venv/lib/python3.6/site-packages/simple_history/models.py", line 386, in <lambda>
    self.history_object, self.history_date
  File "/home/pikachu/Desktop/Practice/library/portal/books/models.py", line 84, in __str__
    return f'[{self.book_id.upper()}] {self.associated_book.title}'
  File "/home/pikachu/Desktop/Practice/library/venv/lib/python3.6/site-packages/django/db/models/fields/related_descriptors.py", line 188, in __get__
    "%s has no %s." % (self.field.model.__name__, self.field.name)
books.models.BookCode.associated_book.RelatedObjectDoesNotExist: BookCode has no associated_book.
```

**Expected behavior**
Model history should be the output.

**Environment:**
 - OS: Ubuntu 18.04.2 LTS
 - Django Simple History Version: 2.7.0
 - Django Version: 2.1.7
 - Database Version: Using default `sqlite3`.

rossmechanic:
[...]

PR #1119, opened 2023-01-31
Add a history_repr field that stores the value of __str__()

alejandro-perez:
Sometimes __str__() requires of a foreign key to represent its value. In those scenarios, two issues appear:

1. Need to perform extra queries. select_related() does not help since the relationship exists in the model, not the historical model.
2. If the related instance is removed, it will fail with DoesNotExist.

One way of overcoming this situation is by storing the value of __str__() at the time we create the Historical Record.

Since this might have a performance penalty, it MUST be explicitly enabled by setting `SIMPLE_HISTORY_HISTORIC_REPR` to True.

## Related Issue
#533

## Motivation and Context
Allowing __str__() that use a ForeignKey to work

## How Has This Been Tested?
It hasn't, it's a pretty simple change.

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -6,6 +6,7 @@ from django.utils import timezone
 from simple_history.utils import (
     get_app_model_primary_key_name,
     get_change_reason_from_object,
+    get_historical_repr,
 )
 
 # when converting a historical record to an instance, this attribute is added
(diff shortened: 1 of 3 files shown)


ddabble:
Thank you for the PR! However, rather than adding another field, I think a better solution is to inform users of how they can customize the string representations of their historical records - potentially only using the historically saved PK of related objects, or removing the related objects from the string altogether. To do this, see [my comment on #1314](https://github.com/jazzband/django-simple-history/issues/1314#issuecomment-1998620572).

(This doesn't directly address the `select_related()` issue you mentioned, but for users determined to have related objects as part of their historical records' string representations, there often exist creative solutions that work around the problem :))

ddabble:
Please reopen if you'd like to discuss alternative solutions 🙂

alejandro-perez:
It's fine. I'm no longer using this module so I'm happy with whatever decision you take.
Thanks!

PR #1128, opened 2023-02-21
Add history diff column to admin change history table

### LINKED ISSUE #886: Default & changed fields
**Problem Statement**
Mostly i changing other fields than primary displayed. When you have more changes (10+) it looks little bit weird - same field. Also current "name" of object is displayed in header

1.  As primary field instead of "Object" set "date/time".
before-after

2. Instead of  "Object" create new field "Changes" and list all fields which were changed for example:
When i change fields: text, summary, description in this new field will be shown (without values) 

Changes  |
------------- |
text, summary, description |

**Additional context**
I  like this package. This is only my opinion. I think it will improve history readability and efficiency when you need roll-back.

### LINKED ISSUE #1030: How to add actions in the admin integration, similar to django default history view
**Problem Statement**
I have changed the is_active column from True to False, now when I look at the history view from admin integration, I am not able to see which field has been changed, but in the default django history list, it shows changes in the action column. How can I do this in the django-simple-history?

In short, I wanted to know, what has been changes by looking at the list display

### LINKED ISSUE #1308: Show diff in history_view
**Problem Statement**
Having a list of historical snapshots with no way to compare them in the admin page is difficult to make comparisons, eg. when tracking down when a particular field changed.

I would like the ability to see the diff between one historical object and the previous one in the history_view in the admin site. I have the code I've written to do this and a screen shot attached below that demonstrate what I'm after. I'd also be very happy to workshop this into a fully integrated solution and submit a merge request for this feature if people think it's a helpful feature.

I've looked for a way to do this with the codebase, and can't find any, I think this is new functionality. It could be done on its own separate page I could build, but this is what the admin site is for. Users could also just manually go through each change to find the one where the value they're looking for changes, but that's tedious and difficult.

**Additional context**
Here's the code I've written to achieve this, I've overridden the history_view method, and inserted this code just after the code that sets the history_list_display values (equivalent to line 72 in simple_history/admin.py)

```
        previous = None
        for list_entry in reversed(action_list):
            if previous:
                diff = list_entry.diff_against(previous).changes
                value = render_to_string("admin/my_app/history_admin_diff.html", {"diff": diff})
            else:
                value = ""
            setattr(list_entry, "diff", value)
            previous = list_entry

        history_list_display = ["diff"] + history_list_display
```

with the template "admin/my_app/history_admin_diff.html" containing this:

```
<table>
    <thead><th>Field</th><th>Old Value</th><th>New Value</th></thead>
    <tbody>
    {% for change in diff %}
    <tr>
        <td>{{ change.field }}</td>
        <td>{{ change.old }}</td>
        <td>{{ change.new }}</td>
    </tr>
    {% endfor %}
    </tbody>
</table>
```

And that creates (for my particular model) a view that looks like this:
image

raunaq-sailo:
## Description

Main changes:
- Added a "Changes" column to `SimpleHistoryAdmin`'s object history table, listing the changes between each historical record of the object; see the docs under "Customizing the History Admin Templates" for overriding its template context (c9963aaeaddd9d66619a9996aa6bb36a70f0d779)
  - *Related: da87859a325a01b256b9890674d0503202e54863, b5e627c70da94a8f960a27a35d6b57c9854e5124, 04eefadf187670b018cbfd3306d5001447eb0c52, 733f4e0a6046917848ef46277079cefcf6c87c98*
- Renamed the (previously internal) admin template `simple_history/_object_history_list.html` to `simple_history/object_history_list.html`, and added the field `SimpleHistoryAdmin.object_history_list_template` for overriding it (16b7de7482d0898c9f83c9042c01442fe35ba66e)
- Deprecated the undocumented template tag `simple_history_admin_list.display_list()`; it will be removed in version 3.8 (16b7de7482d0898c9f83c9042c01442fe35ba66e)
- Added `SimpleHistoryAdmin.get_history_queryset()` for overriding which `QuerySet` is used to list the historical records (18f00bddbfdd0b0d65757982197d670315174ad1)
- Added `SimpleHistoryAdmin.get_history_list_display()` which returns `history_list_display` by default, and made the latter into an actual field (dc3a34eea0a65bbcb3b64865ab500c5896c34e65)
- `ModelDelta` and `ModelChange` (in `simple_history.models`) are now immutable dataclasses; their signatures remain unchanged (28afb4dc8486a7c4563070e25139a107036569b3)
- `ModelDelta`'s `changes` and `changed_fields` are now sorted alphabetically by field name. Also, if `ModelChange` is for an M2M field, its `old` and `new` lists are sorted by the related object. This should help prevent flaky tests. (a842b98a2e36c6ab217831769a1817a28cd59f60)
- `diff_against()` has a new keyword argument, `foreign_keys_are_objs`; see usage in the docs under "History Diffing" (a842b98a2e36c6ab217831769a1817a28cd59f60)
  - *Related: a5439d2d1e274b3b619abacb2700ef6a95f921d0, 0589ac2e637760e5365a023798a7aa0e034aa68f*

## Related Issue

Closes #886, closes #1030, closes #1308.

## Motivation and Context

See the linked issues.

## How Has This Been Tested?

Through the tests added in the changed test files.

image

## Types of changes

## Checklist:


PetrDlouhy:
I tested this PR and it seems to me, that it works correctly and is very helpful. Here is a preview of what it does:
Snímek obrazovky_2023-09-11_12-35-02

I checked, that it does show the diff values for my models and also that it doesn't make unnecessary DB queries (the number of queries is constant and not dependent on number of history records).

When I look at the code, i think there is added unnecessary comment line:
```
<!-- {{ action.history_diff }}-->
```
which is probably some relic from development.

And obviously the failed test would need to be corrected.
@raunaq-sailo Could you please remove the comment and fix the failing tests?

simple_history/templates/simple_history/_object_history_list.html
```diff
@@ -40,6 +41,14 @@
         <td>
             {{ action.history_change_reason }}
         </td>
+        <td>
+          {% if action.history_diff %}
+            {% for diff in action.history_diff %}
+              {{ diff }} <br>
+            {% endfor %}
+<!--            {{ action.history_diff }}-->
+          {% endif %}
```
ddabble, line 50: If my suggestion above is implemented, this can be changed to the following:
```suggestion
          {% if action.history_diff_changes %}
            <ul>
              {% for change in action.history_diff_changes %}
                <li>
                  <code>{{ change.field }}</code>:
                  {{ change.old|truncatechars:100 }}
                  {# Add some spacing, and prevent having the arrow point to the edge of the page if `new` is wrapped #}
                  &nbsp;&rarr;&nbsp;&nbsp;{{ change.new|truncatechars:100 }}
                </li>
              {% endfor %}
            </ul>
          {% endif %}
```
I think the changes would be slightly easier to read if placed inside a `<ul>` list, as suggested above :)

Also note that I added the `truncatechars` filter, in case the old/new values are very long. Not sure if 100 is an "ideal" number of characters, but I think it works.

ddabble (changes requested):
This is very much a neat feature to have! 😄 A few small changes:

ddabble:
@raunaq-sailo Just as a heads-up, I'm considering implementing the changes I suggested myself, if no activity has happened within a couple weeks :)

ddabble (dismissed):

simple_history/admin.py
```diff
@@ -103,6 +111,29 @@ def history_view_title(self, request, obj):
...
             return _("Change history: %s") % force_str(obj)
 
+    def set_history_delta_changes(self, action_list: Sequence[HistoricalChanges]):
+        # Add a `history_delta_changes` attribute to all history records
+        # except the first (oldest) one
+        for i in range(len(action_list) - 1):
+            delta = action_list[i].diff_against(action_list[i + 1])
+            action_list[i].history_delta_changes = [
+                self.format_history_delta_change(change) for change in delta.changes
+            ]
```
tim-schilling, line 121:
```suggestion
        previous = None
        current = None
        for current in action_list:
            if previous is None:
                previous = current
                continue
            delta = previous.diff_against(current)
            previous.history_delta_changes = [
                self.format_history_delta_change(change)
                for change in delta.changes
            ]
```

This allows us to iterate `action_list` just once rather than technically twice with several index operations.

tim-schilling (commented):
I think this is fine, but I would like to raise a slight change in implementation. Rather than always showing the change at the end, would it make more sense to include this as a default option in `history_list_display`? That way, it would allow the user to decide if they want to even include the changes or move it to a different column?

One of the problems with doing that is that we'd have to define some name for this column that would shadow potential fields on the actual historical model. Maybe something like `Historical changes`? I'm thinking `Changes` by itself might be too generic.

I don't feel super strong about it, but it seems like a reasonable customization we could implement. And it would avoid having an explicit call to `set_history_delta_changes`.

tim-schilling:
Oh, I thought `history_list_display` was already a thing. I don't want to have perfect get in the way of good though, so let's make this improvement.

simple_history/admin.py
ddabble: Thanks, that should definitely make it more efficient! I personally think the original version is slightly easier to understand, but your version could make a noticeable difference for large history tables, so let's go with that 🙂

tim-schilling: Is there a particular section of it that could benefit from a comment or two?

ddabble: Not that I can think of 🤔 I guess it's just that I have to pay slightly more attention to what `previous` and `current` are and when they're set and used, but it's such a small algorithm that I think the existing comment (at the top of the method) does the job 🙂

ddabble:
@tim-schilling I realized some additional changes needed to be made to properly display foreign keys and M2M objects, and so most of the recent commits were made to facilitate that - incl. making some optimizations and usage improvements along the way. It's absolutely possible, however, to display them without some of those changes, so let me know if I should split it into multiple PRs 🙂

I also realized that making `history_list_display` contain the current default column layout - like we discussed - would require several extra changes, which I think is better to include in a separate PR that I can open after this one. Is that okay with you?

simple_history/templatetags/simple_history_admin_list.py
```diff
@@ -1,8 +0,0 @@
-from django import template
-
-register = template.Library()
-
-
-@register.inclusion_tag("simple_history/_object_history_list.html", takes_context=True)
-def display_list(context):
```
tim-schilling, line 7: Rather than removing it outright, maybe it should raise a deprecation warning?

simple_history/manager.py
```diff
@@ -91,6 +91,18 @@ def latest_of_each(self):
             )
         return latest_historics
 
+    def select_related_history_tracked_objs(self):
```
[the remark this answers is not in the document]
ddabble: > I'm not sure this is a great idea. I do see the utility in it, but it makes assumptions on the user's application. It's also something that the user should be able to do on their own. Once this gets merged, it becomes very difficult to remove or change to maintain backwards compatibility. It's easier to keep the library slim (if it can event be called that 😁) and provide hooks for the user to do what they need.

Good point, I agree 👍

> If this is necessary for the diffing logic, then I'd prefer if it were an undocumented method so we can change it as needed moving forward.

It's necessary to reduce the number of DB queries, but I see no problem in simply moving the code to [the only place it's currently used](https://github.com/raunaq-sailo/django-simple-history/blob/d054c33e75e32851cefc59c19ff2225e26d2808d/simple_history/admin.py#L132). Not sure whether I prefer removing the documentation or inlining the code in the previously linked method 🤔 Any opinions on those two alternatives?

simple_history/templatetags/simple_history_admin_list.py
ddabble: Hm yeah, it didn't occur to me that some users might be using this for overriding parts of the template(s) - if that's what you were thinking about - so I agree that we should deprecate it first.

ddabble: I don't have a lot of experience deprecating things in Python/Django libraries, so let me know if you're familiar with any technique that would be preferred over what I added in 8513231192aba4e52d51bbe708dc716f162e6a3a 🙂 Also, any opinions on whether 3.8 sounds like a good version to remove the template tag?

tim-schilling: I think if our project had a regular release cadence we could have a formal deprecation policy. I think what you'd laid out works perfectly well!

simple_history/manager.py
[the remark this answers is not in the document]
tim-schilling: I like the separate function. It gets us closer to something similar to https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_select_related, but I don't feel strongly about it.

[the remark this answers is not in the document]
ddabble: True 👍
I'll prefix the method with `_` while I'm at it.

ddabble:
@tim-schilling I mainly added some (presumably final) polish commits, including fd72e7a5f9943a3d93be4066a97ed6d946b0c4d8 and 38ec0c2049bc2388408d0701a69cefed1bbad45d, which improve how really long strings are displayed, and add tests for how safe strings are handled. Again, please let me know if I should split it into multiple PRs 🙂

(The failing tests are unrelated to this PR; I'm looking into them.)

tim-schilling:
@ddabble I think it looks good. There was a lot in those changes and admittedly I didn't do a great job of reviewing. It looks like the code is well written, there are tests and they are passing. Not having code coverage hurts us here because that would confirm that the code is actually running in the tests.

Let's make this the last of the large, multi-commit PRs though. They're really tough to do a good job when I'm not super-duper familiar with the codebase.

tim-schilling (dismissed):

tim-schilling:
@ddabble when you think it's ready, I'll integrate it with our application to see how things perform.

simple_history/template_utils.py
```diff
@@ -99,23 +117,46 @@ def prepare_delta_change_value(
...
-        if isinstance(value, list):
-            value = f'[{", ".join(map(str, value))}]'
 
-        value = conditional_escape(value)
-        value = truncatechars_html(value, self.max_displayed_delta_change_chars)
-        return value
+        def stringify_value(value) -> Union[str, SafeString]:
+            # If `value` is a list, stringify each element using `str()` instead of
+            # `repr()` (the latter is the default when calling `list.__str__()`)
+            if isinstance(value, list):
```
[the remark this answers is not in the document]
ddabble: I had the same thought when writing it, but decided to only support lists to prevent unexpected behavior, since e.g. strings are instances of `Sequence`.

`value` can be of any type (returned by `prepare_delta_change_value()`), and the thought was that only M2M fields will use this code by default (since their values are lists), and that users wanting to leverage `stringify_value()`'s code could override `prepare_delta_change_value()` and return e.g. a list of objects whose `__str__()` implementation looks good enough for being displayed in the template through the `f"[{...}]"` string. *(I guess the latter point should ideally be added to the docstrings 😅)*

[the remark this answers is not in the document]
ddabble: With that in mind, do you have any alternative implementation suggestions..? I'm not sure I can think of any without making the code unnecessarily complicated 🤔

ddabble:
> Not having code coverage hurts us here because that would confirm that the code is actually running in the tests.

True.. From the workflow logs, it looks like we're being rate-limited by Codecov, since the repo doesn't have any upload token 😕 (see #1305)

> Let's make this the last of the large, multi-commit PRs though. They're really tough to do a good job when I'm not super-duper familiar with the codebase.

Alright, that's fair 😅 Thanks a bunch for reviewing regardless! I'll clean up the commits for merging, as mentioned above (followed by re-requesting a review) 🙂

> when you think it's ready, I'll integrate it with our application to see how things perform.

I'm assuming you mean as a final system test before merging..?

simple_history/template_utils.py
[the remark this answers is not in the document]
tim-schilling: Nope, that's a good enough justification to me!

tim-schilling:
> I'm assuming you mean as a final system test before merging..?

Yeah, that's what I was thinking. Let me do that now.

tim-schilling:
Also, I didn't run into any issues. All good.

ddabble:
Finished cleaning up the commits, so the PR is ready for merging :)
*([A few minor changes done while rebasing](https://github.com/jazzband/django-simple-history/compare/5a772dd71c37311ba3b5d08abc6d8a24e93016d2..733f4e0a6046917848ef46277079cefcf6c87c98))*

tim-schilling (approved):
This looks good. I really appreciate your attention to the docs, code comments and organization.

hfroot:
Not to spam, but I think I love you 😍 this feature is awesome

Issue #1152, opened 2023-03-28
With HistoricForeignKey prefetch_related doesn't return all related objects in some situations


duebbert:
**Describe the bug**

There is an issue when using `HistoricForeignKey` and a "bare" `prefetch_related`. It doesn't return all results when calling the related set, for example `self.some_set.all()`

This happens when doing a `.prefetch_related("some_set")` in the queryset.

It does *NOT* happen when:
1. doing a `.prefetch_related(Prefetch("some_set", queryset=Some.objects.all()))`
2. or not doing a prefetch_related (which obviously leads to N+1 then)
3. or changing back to the normal `models.ForeignKey`.

This seems to happen when prefetching related objects more than one level down.

**Reproduction**

Models:
```python
from django.db import models
from simple_history.models import HistoricForeignKey, HistoricalRecords


class Main(models.Model):
    name = models.CharField(max_length=100)
    history = HistoricalRecords()


class Sub(models.Model):
    main = HistoricForeignKey(Main, on_delete=models.CASCADE)
    name = models.CharField(max_length=100)
    history = HistoricalRecords()

    def __str__(self):
        return f"linked to {self.main} ({self.pk})"


class SubSub(models.Model):
    sub = HistoricForeignKey(Sub, on_delete=models.CASCADE)
    name = models.CharField(max_length=100)
    history = HistoricalRecords()

    def __str__(self):
        return f"linked to {self.sub} ({self.pk})"
```

Code:
```python
from testing.models import Main, Sub, SubSub
from django.db.models import Prefetch

main = Main.objects.create(name="Main 1")

sub1 = Sub.objects.create(name="Sub 1", main=main)
sub2 = Sub.objects.create(name="Sub 2", main=main)

subsub1 = SubSub.objects.create(name="SubSub 1", sub=sub1)
subsub2 = SubSub.objects.create(name="SubSub 2", sub=sub2)

# Correct behaviour
main = Main.objects.get(name="Main 1")
for sub in main.sub_set.all():
    print(sub, sub.subsub_set.all())

# Correct behaviour
main = Main.objects.prefetch_related("sub_set").get(name="Main 1")
for sub in main.sub_set.all():
    print(sub, sub.subsub_set.all())

# Correct behaviour
main = Main.objects.prefetch_related(
    Prefetch("sub_set__subsub_set", queryset=SubSub.objects.all())
).get(name="Main 1")
for sub in main.sub_set.all():
    print(sub, sub.subsub_set.all())

[...]

PR #1159, opened 2023-04-14
Fix HistoricForeignKey when used together with prefetch_related()

mr-mare:
## Description
This fixes behaviour of `HistoricForeignKey` when used together with `prefetch_related()`.

The following is sufficient to reproduce the bug:
- models:
```python
from django.db import models
from simple_history.models import HistoricalRecords, HistoricForeignKey


class Poll(models.Model):
    question = models.CharField(max_length=200)

    history = HistoricalRecords()


class Choice(models.Model):
    poll = HistoricForeignKey(Poll, on_delete=models.CASCADE)
    choice = models.CharField(max_length=200)
    votes = models.IntegerField()

    history = HistoricalRecords()
```
- code:
```python
from poll.models import Poll, Choice

why_poll = Poll.objects.create(question="why?")
how_poll = Poll.objects.create(question="how?")

Choice.objects.create(poll=why_poll, votes=0)
Choice.objects.create(poll=how_poll, votes=0)

for poll in Poll.objects.prefetch_related("choice_set").all():
    print(poll.choice_set.all())

# expected result:
# <QuerySet [<Choice: Choice object (1)>]>
# <QuerySet [<Choice: Choice object (2)>]>

# actual result:
# <QuerySet [<Choice: Choice object (1)>]>
# <QuerySet []>
```

## Related Issue
#1152 

## Motivation and Context
Property `related_manager_cls` is defined in [`HistoricReverseManyToOneDescriptor`](../blob/3.3.0/simple_history/models.py#L827). This property returns a function call of [`create_reverse_many_to_one_manager`](https://github.com/django/django/blob/4.2/django/db/models/fields/related_descriptors.py#L632) function and passes it `HistoricRelationModelManager` as an argument. However, `create_reverse_many_to_one_manager` defines `RelatedManager`, which subclasses the manager passed as an argument and returns this `RelatedManager` class.

When `prefetch_related` is used, [`get_prefetch_queryset`](https://github.com/django/django/blob/4.2/django/db/models/fields/related_descriptors.py#L730) method of `RelatedManager` is called.
The problem lies in the fact that it bypasses `RelatedManager.get_queryset` method which does filtering to instance (`_apply_rel_filters`) by calling `super().get_queryset()`. But due to inheritance, it gets evaluated to `HistoricRelationModelManager.get_queryset` which does the same filtering as well.

This SQL is generated then (notice the `"poll_choice"."poll_id" = 1` condition):
```sql
SELECT "poll_poll"."id", "poll_poll"."question" FROM "poll_poll"
SELECT "poll_choice"."id", "poll_choice"."poll_id", "poll_choice"."choice", "poll_choice"."votes" FROM "poll_choice" WHERE ("poll_choice"."poll_id" = 1 AND "poll_choice"."poll_id" IN (1, 2))
```

Removing `_apply_rel_filters` call from `HistoricRelationModelManager.get_queryset` solves this problem.

On the other hand, since `create_reverse_many_to_one_manager` returns `RelatedManager`, rather than `HistoricRelationModelManager`, the only way to access `HistoricRelationModelManager.get_queryset` is by calling `super().get_queryset()` from `RelatedManager`. This call can be found:
- in `RelatedManager.get_prefetch_queryset` where it is undesired,
- in `RelatedManager.get_queryset` where `_apply_rel_filters` is called right after `super().get_queryset()`.

Therefore, I believe it is safe to remove it.

## How Has This Been Tested?
Added new tests.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changes
 Unreleased
 ----------
 
+- Fixed ``HistoricForeignKey`` behaviour when used together with ``prefetch_related()``
 
 3.3.0 (2023-03-08)
 ------------------
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -860,7 +860,7 @@ class HistoricReverseManyToOneDescriptor(ReverseManyToOneDescriptor):
                         )
                     else:
                         queryset = super().get_queryset()
-                    return self._apply_rel_filters(queryset)
+                    return queryset
 
         return create_reverse_many_to_one_manager(
             HistoricRelationModelManager, self.rel
--- a/simple_history/tests/tests/test_models.py
+++ b/simple_history/tests/tests/test_models.py
@@ -2545,3 +2545,96 @@ class HistoricForeignKeyTest(TestCase):
         )[0]
         pt1i = pt1h.instance
         self.assertEqual(pt1i.organization.name, "original")
+
+    def test_non_historic_to_historic_prefetch(self):
+        org1 = TestOrganizationWithHistory.objects.create(name="org1")
+        org2 = TestOrganizationWithHistory.objects.create(name="org2")
+
+        p1 = TestParticipantToHistoricOrganization.objects.create(
+            name="p1", organization=org1
+        )
+        p2 = TestParticipantToHistoricOrganization.objects.create(
+            name="p2", organization=org1
+        )
+        p3 = TestParticipantToHistoricOrganization.objects.create(
+            name="p3", organization=org2
+        )
+        p4 = TestParticipantToHistoricOrganization.objects.create(
+            name="p4", organization=org2
+        )
+
+        with self.assertNumQueries(2):
+            record1, record2 = TestOrganizationWithHistory.objects.prefetch_related(
+                "participants"
+            ).all()
+
+            self.assertListEqual(
+                [p.name for p in record1.participants.all()],
+                [p1.name, p2.name],
+            )
+            self.assertListEqual(
+                [p.name for p in record2.participants.all()],
+                [p3.name, p4.name],
+            )
+
+    def test_historic_to_non_historic_prefetch(self):
+        org1 = TestOrganization.objects.create(name="org1")
+        org2 = TestOrganization.objects.create(name="org2")
+
+        p1 = TestHistoricParticipantToOrganization.objects.create(
+            name="p1", organization=org1
+        )
+        p2 = TestHistoricParticipantToOrganization.objects.create(
+            name="p2", organization=org1
+        )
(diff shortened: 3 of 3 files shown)


evanhaldane:
Plans to review/merge this?

(I'm in the process of integrating this package for the first time and I spent a few hours chasing around a bug until I realized it was caused by this issue.)

cc: @ddabble ?

anyidea:
any progress?

robertofd-nextmatter:
Any reason why this is not getting merged? We are also having the same issue

mr-mare:
I'm happy to help. If there is anything I can do, please let me know.

ddabble (changes requested):
Nice, good job investigating this solution! 😄 I agree with your reasoning for why it should be safe removing the call to `_apply_rel_filters()`. Furthermore, it looks like `create_reverse_many_to_one_manager()` usually gets a standard `Manager` as its `superclass` arg, which doesn't contain any `_apply_rel_filters()` method. And so it seems likely that `RelatedManager` was designed to be expecting that its superclass' `get_queryset()` method doesn't do any of the things that `_apply_rel_filters()` does, which again supports the decision to remove the call.

Do you think we should remove the `try` block with `self.instance._prefetched_objects_cache[...]` as well? The exact same call is in `RelatedManager.get_queryset()`, but it doesn't seem to hurt performance much... Though I'm not sure about the consequences of removing it (or any unknown, existing bugs that might live on if we leave it there, for that matter), considering that `RelatedManager.get_prefetch_querysets()` seems to skip that cache lookup by calling `super().get_queryset()`, combined with the fact that the mentioned standard `Manager` also doesn't make any cache lookups in `get_queryset()` 🤔

Lastly, could you add a test case for a nested `prefetch_related()` call, like the `SubSub` examples in the description of #1152? 🙂 I guess it should be enough e.g. adding a new history-tracked model with a `HistoricForeignKey` referencing `TestHistoricParticipanToHistoricOrganization`, and test with the different arguments to `prefetch_related()` from the mentioned `SubSub` examples, which would include `"historic_participants__<subsub related_name>"` and various nested variants using `Prefetch(...)`.

Other than that, very clear and thorough test cases! And sorry that it took so long getting to reviewing this!

tomhampshire:
Hi @mr-mare, hope you don't mind: I've opened #1604, which is your commit cherry-picked onto current master (your authorship preserved) plus two follow-up commits addressing @ddabble's review (the cache-lookup removal, and a nested prefetch regression test for the SubSub case from #1152). Closing-or-keeping this PR is up to you and the maintainers; please feel free to take ownership of #1604 or reuse anything from it as you see fit.

ddabble:
Superseded by #1604

PR #1188, opened 2023-06-12
use new style middleware and catch exceptions to be sure to clear request in context

### LINKED ISSUE #1189: `HistoryRequestMiddleware` doesn't clean up after itself under some circumstances
**Describe the bug**
We caught a problem were the request was reuse, causing a strange random behaviour.

**To Reproduce**
- a test makes a request, so the context is updated
- an exception happens and process_response is never called, context still contains WRONG request
- the next test creates some versioned model, and the WRONG request is used

**Expected behavior**
Simple history not using the previous request.

 - OS: [docker python:3.8 image]
 - Django Simple History Version: 3.3.0
 - Django Version: 2.2.
 - Database Version: MySQL 8.0

Pull request here: https://github.com/jazzband/django-simple-history/pull/1188

Will complete the checklist if you think this should go in.

manelclos:
## Description
We caught a problem were the request was reused, causing a strange random behaviour:

- a test makes a request, so the context is updated
- an exception happens and process_response is never called, context still contains WRONG request
- the next test creates some versioned model, and the WRONG request is used

## Related Issue
Resolves #1189.

## How Has This Been Tested?
Extensive pytest codebase using django-simple-history
Fixes the problem we were facing

## Types of changes

## Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,8 @@ Unreleased
 - Dropped support for Django 4.0, which reached end-of-life on 2023-04-01 (gh-1202)
 - Added support for Django 4.2 (gh-1202)
 - Made ``bulk_update_with_history()`` return the number of model rows updated (gh-1206)
+- Fixed ``HistoryRequestMiddleware`` not cleaning up after itself (i.e. deleting
+  ``HistoricalRecords.context.request``) under some circumstances (gh-1188)
 
 3.3.0 (2023-03-08)
 ------------------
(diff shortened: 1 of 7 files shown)


ddabble:
@manelclos I added an exception variable to your code, in addition to adding a test and updating the documentation 🙂

ddabble (dismissed):
Great, thank you! 😊

Do review the changes I pushed, if you have any feedback :) Otherwise, I'll merge in a week's time.

manelclos:
Hi @ddabble all good! I didn't notice the error was documented already. Thanks for adding the tests and documentation.

Issue #1174, opened 2023-05-04
How to specify table name format in base model depending on child model's overriden table name ?


avinashs2401:
Hi,
I have a Base model like this, which many models would inherit from,

```
class BaseModel(models.Model):
    created_date = models.DateTimeField(auto_now_add=True)
    last_updated_date = models.DateTimeField(auto_now=True)
    history = HistoricalRecords(inherit=True)
```

I have child models inheriting from this BaseModel and I have overriden the db table names to be with underscores. Eg.
```
class ProductVariants(BaseModel):
   title = models.CharField(max_length=255, blank=False, null=False)
   class Meta:
       db_table = "product_variants"
```

I want History table to be created as `appname_historical_product_variants`. But right now its getting created as `appname_historicalproductvariants` 

Any suggestions on how to achieve this?

singhravi1:
@avinashs2401   
You can override `HistoricalRecords` class and add an attribute,  
```
DEFAULT_MODEL_NAME_PREFIX = "Historical_"
```

Now use this new class in your models for the history field.

singhravi1:
@avinashs2401 
You can use `HistoricalRecords` attribute `table_name` to mention the table name for the historical model.

singhravi1:
@avinashs2401 
If you manage it in the base model only, you can override the `HistoricalRecords` class and specify a custom attribute `DEFAULT_MODEL_NAME_PREFIX`.
Otherwise you can mention the `history` field specifically each model with `table_name` attribute.

muneeb706:
@avinashs2401 
As per my understanding, there is no way to specify custom table name for history models of inherited child models.
You can only change the history table name where HistoricalRecords instance is created (i.e. in the Base Class).

Overriding db_table property of Meta will only update table name for the model itself ( this is done by django: https://docs.djangoproject.com/en/4.2/ref/models/options/#db-table).

singhravi1:
@muneeb706 
You could override the `HistoricalRecords`'s `create_history_model` method and add a logic using the model name being generated, to change the `attrs["Meta"].db_table` as per your requirements

muneeb706:
@singhravi1 

Do you mean, something like this ?

```
from django.db import models
from simple_history.models import HistoricalRecords


class CustomHistoricalRecords(HistoricalRecords):
    def create_history_model(self, model, inherited):
        history_model = super().create_history_model(model, inherited)
        history_model._meta.db_table = "historical_" + model._meta.db_table
[...]

PR #1196, opened 2023-06-24
Fix: Error when setting table name with inherit

### LINKED ISSUE #1195: Error when setting table_name with inherit=True
**Problem Statement**

When `table_name` is set with `inherit=True` in base model then error is thrown when `makemigrations` command is entered.

***For example:***

```
from django.db import models
from simple_history.models import HistoricalRecords

class BaseModel(models.Model):
    created_date = models.DateTimeField(auto_now_add=True)
    last_updated_date = models.DateTimeField(auto_now=True)
    history = HistoricalRecords(inherit=True, table_name='base_model_history')
  
class ChildModel(BaseModel):
    title = models.CharField(max_length=255, blank=False, null=False)

```    
***Following error is thrown with above code:***

(models.E028) db_table 'base_model_history' is used by multiple models: inherit.HistoricalBaseModel, inherit.HistoricalChildModel.

***Cause of Problem***

As per my understanding, after looking at the source code, I found that `Meta.db_table` property of model is updated when `table_name` property is set. In case, when `inherit=True`, `self.table_name` still points to `table_name ` set in the base model when `create_history_model `method is invoked for the child model. This logic is implemented in line 301-302 in [create_history_model ](https://github.com/jazzband/django-simple-history/blob/master/simple_history/models.py) method.

***Possible Solution***

inherited flag is sent as an argument to `create_history_model` method, this can be used in the condition where table_name is checked. Condition at line 301 will look like this:
`if not inherited and self.table_name is not None:`

If it can not be done, then this behavior should be mentioned explicitly in the documentation.

Overridden `create_history_model` method to update `table_name` when inherited is set to True and table_name is set:

```
from django.db import models
from simple_history.models import HistoricalRecords


class CustomHistoricalRecords(HistoricalRecords):
    def create_history_model(self, model, inherited):
        if inherited and self.table_name:
            self.table_name=f"{model._meta.app_label}_{self.get_history_model_name(model)}"
        history_model = super().create_history_model(model, inherited)
        return history_model

class BaseModel(models.Model):
    created_date = models.DateTimeField(auto_now_add=True)
    last_updated_date = models.DateTimeField(auto_now=True)
    history = CustomHistoricalRecords(inherit=True, table_name="base_model_history")
  
class ChildModel(BaseModel):
    title = models.CharField(max_length=255, blank=False, null=False)
```

**Additional context**

I found this issue while looking for the solution of  #1174

muneeb706:
Error was thrown when `table_name` is set with inherit=True in base model.
inherited check added in the condition where `table_name` is checked for updating db table name. 

## Description

inherited flag argument in `create_history_model` method is used in the condition where `table_name` is checked. Condition at line 301 of `simple_history.models` now look like this:
`if not inherited and self.table_name is not None:`

## Related Issue
Closes #1195.

## Motivation and Context
I found this issue while looking for the solution of #1174.
In documentation it is not mentioned that table_name can't be used with inherit. It is implied that this should work, this inconsistency was the motivation.

## How Has This Been Tested?
Ran tests locally using `runtests.py`  with python (3.8,3.11) and django 4.1

## Types of changes

## Checklist:


theryanwalker (commented):
Looks good to me! It would be great to have a test for this fix.

muneeb706:
> Looks good to me! It would be great to have a test for this fix.

Thank you for reviewing, I have added a test.

theryanwalker (approved):
LGTM, thanks for the test @muneeb706!

ddabble (approved):
Looks good! Thank you, @muneeb706, and also @theryanwalker for reviewing :)

PR #1197, opened 2023-06-24
Assert table name inherit options

singhravi1:
Mentioning `table_name` attribute would be invalid with `inherit=True` as it will cause multiple tables to have same table name and that error does not clearly specify this issue scenario. Therefore added an informed error message.

## Description
Added check if `inherit=True` and `table_name`, both options are passed and raised a `ValueError`.

## Motivation and Context
Informing clearly about the issue to the user as they may get confused with the `MultipleRegistrationsError` raised.

## How Has This Been Tested?
Added a test `test_inherited_static_table_name_abstract_base` and used `runtest` to check its validity.

## Types of changes

## Checklist:

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -106,6 +106,7 @@ Authors
 - Phillip Marshall
 - Prakash Venkatraman (`dopatraman <https://github.com/dopatraman>`_)
 - Rajesh Pappula
+- Ravi Singh (`singhravi1 <https://github.com/singhravi1>`_)
 - Ray Logel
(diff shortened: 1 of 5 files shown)


ddabble:
#1196 just made it legal to pass a `table_name` when `inherit=True`, by making child models not inherit `table_name`. So since this PR essentially adds a clearer error message for the previous behavior, it's no longer necessary and can be closed. Do post a comment if you disagree :)

In any case, thank you so much for the contribution! 😊

PR #1256, opened 2023-09-15
Handle missing request

Tenzer:
## Description
The change in https://github.com/jazzband/django-simple-history/pull/1188 removed a check for if the request context had a "request" entry in it before trying to delete the request data.

## Motivation and Context
This has a side effect of breaking tests in order projects in very specific circumstances, where multiple requests pass through the middleware, concurrently, inside the same thread, as it means the middleware then may try to remove the request key on a context where it already has been removed.

In our case we make us of [before_after](https://github.com/c-oreills/before_after) to test some behaviours that are specific to concurrent requests, which allows us to execute concurrent requests inside one test. In those cases the test have started failing with django-simple-history 3.4.0 because an `AttributeError` is raised when the second request passes back up through the middleware, and tries to delete the request data after it already has been done as part of the first request.

This shouldn't have any impact in real-life scenarios.

## How Has This Been Tested?
I have run our test suite (for proprietary code) before and after this change to verify it no longer raises the `AttributeError` and our test suite passes.

## Types of changes

## Checklist:

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -64,6 +64,7 @@ Authors
 - Jake Howard (`RealOrangeOne <https://github.com/realorangeone>`_)
 - James Muranga (`jamesmura <https://github.com/jamesmura>`_)
 - James Pulec
+- Jeppe Fihl-Pearson (`Tenzer <https://github.com/Tenzer>`_)
 - Jesse Shapiro
 - Jihoon Baek (`jihoon796 <https://github.com/jihoon796>`_)
 - Jim Gomez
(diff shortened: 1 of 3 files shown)


ddabble (approved):
Looks good, thank you! 😊

Tenzer:
Any idea when a new release might be cut that includes this update?

ddabble:
@Tenzer This is now included in version 3.5.0 :slightly_smiling_face:

Tenzer:
Great, thanks! I noticed earlier today and have upgraded to the latest version already.

Issue #642, opened 2020-04-13
Best way to dynamically disable creation of history records?


megavenik:
**Problem Statement**
For example, I have some kind of blog, where I do track the history of users' comments. And when a user deletes his comment under any blog post I still want to be able to see it, and maybe recover back.

But when I delete that post, I don't need to know anything about any of those comments any more. And I want not only to remove the existing history (it's quite an easy task) but also **not to create** "-" historical records caused by cascade delete.

**Describe the solution you'd like**
The most user-friendly way to do this is to use some method like "save_without_historical_record", described is the docs.

**Describe alternatives you've considered**
As another possibility, maybe you could tell which Signal I need to disconnect before I call Post.object.delete() ? And how to perform it properly. I've tried to find any help reading the library source code, but, seem to be failed:(

rossmechanic:
Hi @megavenik – I can't think of a way to disconnect the post_delete signal, as there's no easy way to access the `HistoricalRecords` model instance that has the receiver function in the first place. Doesn't seem like there's an easy way to pick when you want to do this, although the `cascade_delete_history` does what you want (you just can't choose when to turn it on or off dynamically). Feels like your best bet is to put up a PR that allows you to specify the `cascade_delete_history` attribute dynamically, similarly to how it's done on `skip_history_when_saving`. Referencing to the two sections of the code I'm referring to below:

https://github.com/treyhunner/django-simple-history/blob/master/simple_history/models.py#L472
https://github.com/treyhunner/django-simple-history/blob/master/simple_history/models.py#L466

jeking3:
I added this to my model base class, it's quite effective:
```
    @contextmanager
    def no_history(self):
        """
        Disable history generation for this instance.
        """
        assert not hasattr(  # nosec
            self, "skip_history_when_saving"
        ), "it is not safe to nest no_history contexts"
        self.skip_history_when_saving = True
        try:
            yield
        finally:
            del self.skip_history_when_saving
```

If this solves your issue, please consider closing it.  Thanks.

PR #1262, opened 2023-09-26
Make `skip_history_when_saving` work when creating a new object

### LINKED ISSUE #1141: skip_history_when_saving does not work as expected
**Describe the bug**
I am trying to avoid writing to history when certain condition is met. I have a base model that looks the following.
```python
class BaseModel(models.Model):
    ...
    
    def save(self, *, skip_history_when_saving=True, **kwargs):
        if skip_history_when_saving:
            self.skip_history_when_saving = skip_history_when_saving
        super().save(**kwargs)
```
For my models, I am creating it using:
```python
class MyModel(BaseModel):
    ...

my_model = MyModel()
my_model.full_clean()
my_models.save(skip_history_when_saving=True) #expecting only one insertion into the MyModel db but seeing an entry being created for simple history MyModel table as well)
```
`save()` gets called as expected in the base model, at the end of execution, it's still creating an entry in historical table. I can confirm it by looking at the sql query. 

**Expected behavior**
Expecting only the model instance to be created without creating the historical model instance on post creation. 

 - Django Simple History Version: [e.g. 3.2.0]
 - Django Version: [e.g. 4.1.7]

ddabble:
## Description

`skip_history_when_saving` used to only apply when updating an existing object (due to [the `not created` operand in `post_save()`](https://github.com/jazzband/django-simple-history/commit/3c2319c4f9164fb51152e94a03c27f2d978cae9e#diff-a13c48090526a77430e6819c61ffb009b8ed15ffca1b40f3eef675c8c9b77db9L654)), not when creating a new one. This PR fixes that.

Also improved the docs on saving without creating historical records.

## Related Issue

Closes #1141.

## Motivation and Context

Fixing the linked issue.

## How Has This Been Tested?

See the test added in `test_models.py`.

## Types of changes

## Checklist:

### DIFF
--- a/docs/querying_history.rst
+++ b/docs/querying_history.rst
@@ -175,12 +175,11 @@ model history.
     <Poll: Poll object as of 2010-10-25 18:04:13.814128>
 
 
-Save Without Creating Historical Records
+Save without creating historical records
 ----------------------------------------
 
 If you want to save model objects without triggering the creation of any historical
-records, you can set a ``skip_history_when_saving`` attribute to ``True``
-on each object before saving - for example like this:
+records, you can do the following:
 
 .. code-block:: python
 
@@ -189,7 +188,7 @@ on each object before saving - for example like this:
     # We recommend deleting the attribute afterward
     del poll.skip_history_when_saving
 
-This also works when creating an object, but only when calling the ``save`` method:
+This also works when creating an object, but only when calling ``save()``:
 
 .. code-block:: python
 
@@ -198,6 +197,9 @@ This also works when creating an object, but only when calling the ``save`` meth
     poll.skip_history_when_saving = True
     poll.save()
     del poll.skip_history_when_saving
+
+.. note::
+    Historical records will always be created when calling the ``create()`` manager method.
 
 Alternatively, call the ``save_without_historical_record()`` method on each object
 instead of ``save()``.
--- a/docs/querying_history.rst
+++ b/docs/querying_history.rst
@@ -189,8 +189,7 @@ on each object before saving - for example like this:
     # We recommend deleting the attribute afterward
     del poll.skip_history_when_saving
 
-This also works when creating an object
-(albeit only when calling the model's standard Python constructor followed by saving):
+This also works when creating an object, but only when calling the ``save`` method:
 
 .. code-block:: python
 
@@ -216,7 +215,7 @@ and it looks like this:
             del self.skip_history_when_saving
         return ret
 
-As a final option, you can disable the creation of historical records for *all* models
+Or disable the creation of historical records for *all* models
 by adding the following line to your settings:
 
 .. code-block:: python
--- a/docs/querying_history.rst
+++ b/docs/querying_history.rst
@@ -175,30 +175,49 @@ model history.
     <Poll: Poll object as of 2010-10-25 18:04:13.814128>
 
 
-Save without a historical record
---------------------------------
+Save Without Creating Historical Records
+----------------------------------------
 
-If you want to save a model without a historical record, you can use the following:
+If you want to save model objects without triggering the creation of any historical
+records, you can set a ``skip_history_when_saving`` attribute to ``True``
+on each object before saving - for example like this:
 
 .. code-block:: python
 
-    class Poll(models.Model):
-        question = models.CharField(max_length=200)
-        history = HistoricalRecords()
+    poll.skip_history_when_saving = True
+    poll.save()
+    # We recommend deleting the attribute afterward
+    del poll.skip_history_when_saving
 
-        def save_without_historical_record(self, *args, **kwargs):
-            self.skip_history_when_saving = True
-            try:
-                ret = self.save(*args, **kwargs)
-            finally:
-                del self.skip_history_when_saving
-            return ret
+This also works when creating an object
+(albeit only when calling the model's standard Python constructor followed by saving):
 
+.. code-block:: python
 
-    poll = Poll(question='something')
-    poll.save_without_historical_record()
+    # Note that `Poll.objects.create()` is not called
+    poll = Poll(question="Why?")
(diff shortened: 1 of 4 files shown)


docs/querying_history.rst
```diff
@@ -175,30 +175,49 @@ model history.
     <Poll: Poll object as of 2010-10-25 18:04:13.814128>
 
 
-Save without a historical record
---------------------------------
+Save Without Creating Historical Records
```
tim-schilling, line 178: Most titles don't use title casing in the file. I think you should reconsider this choice.

docs/querying_history.rst
```diff
@@ -175,30 +175,49 @@ model history.
...
+----------------------------------------
 
-If you want to save a model without a historical record, you can use the following:
+If you want to save model objects without triggering the creation of any historical
+records, you can set a ``skip_history_when_saving`` attribute to ``True``
+on each object before saving - for example like this:
```
tim-schilling, line 183: Personally, I think the previous version is better. It's fewer words and you have a code example to show the how.

docs/querying_history.rst
```diff
@@ -175,30 +175,49 @@ model history.
...
 
-        def save_without_historical_record(self, *args, **kwargs):
-            self.skip_history_when_saving = True
-            try:
-                ret = self.save(*args, **kwargs)
-            finally:
-                del self.skip_history_when_saving
-            return ret
+This also works when creating an object
+(albeit only when calling the model's standard Python constructor followed by saving):
```
tim-schilling, line 193:
```suggestion
This also works when creating an object, but only when calling the ``save`` method:
```

I think it simplifies the language a bit. I'm not tied particularly strongly to it.

docs/querying_history.rst
```diff
@@ -175,30 +175,49 @@ model history.
...
+
+.. code-block:: python
 
+    # Note that `Poll.objects.create()` is not called
```
tim-schilling, line 197: Maybe we put this after the code example?

> Historical records will always be created using the ``create`` manager method.

docs/querying_history.rst
```diff
@@ -175,30 +175,49 @@ model history.
...
+    def save_without_historical_record(self, *args, **kwargs):
+        self.skip_history_when_saving = True
+        try:
+            ret = self.save(*args, **kwargs)
+        finally:
+            del self.skip_history_when_saving
+        return ret
 
-Or disable history records for all models by putting following lines in your ``settings.py`` file:
+As a final option, you can disable the creation of historical records for *all* models
```
tim-schilling, line 219:
```suggestion
Or disable the creation of historical records for *all* models
```

I think the previous version was a bit too wordy for technical documentation.

simple_history/models.py
```diff
@@ -180,9 +180,9 @@ def contribute_to_class(self, cls, name):
     def add_extra_methods(self, cls):
         def save_without_historical_record(self, *args, **kwargs):
             """
-            Save model without saving a historical record
+            Save the model instance without creating a historical record.
 
-            Make sure you know what you're doing before you use this method.
+            *Make sure you know what you're doing before using this method.*
```
tim-schilling, line 185: nit: I'm not a fan of trying to use markup here.

docs/querying_history.rst
tim-schilling: Also I bet it's possible for us to manage this by overriding the ``create`` method on the manager class and accepting ``skip_history_when_saving`` as a kwarg.

tim-schilling (changes requested):
If you disagree with my review, I could be convinced to leave things as is.

docs/querying_history.rst
ddabble: Ah yes, I agree 👌 I was planning on polishing the docs in a later PR, which includes title-casing all the headings - to match both the main headings in the docs sidebar and various subheadings on some of the pages.

Any opinions on whether we should make title casing or normal casing the norm for the subheadings, btw?

docs/querying_history.rst
ddabble: Good point on already having a code example 😅 I'll push a slightly shorter version 🙂 I do prefer being specific about "triggering the creation", though.

docs/querying_history.rst
ddabble: Sure! `save()` is what triggers our signal handlers, after all 👌

docs/querying_history.rst
ddabble: > Maybe we put this after the code example?
> > Historical records will always be created using the ``create`` manager method.

Sure, I can place it inside a `.. note::` :)

docs/querying_history.rst
ddabble: Fair, I'll add it :)

simple_history/models.py
ddabble: We do have somewhat similar markup in other docstrings (like double instead of single backticks) that's useful when rendering them with previewing tools in some IDEs 🤷 Though I do see that it doesn't add that much value in this case, and it looks slightly out of place when reading the code directly, so I'll remove it 🙂

docs/querying_history.rst
tim-schilling: I prefer the normal casing for sentence titles here, but it's not something I'd argue to change an existing set of documentation over to.

docs/querying_history.rst
tim-schilling: > You think we should do that in this PR?

Definitely no :smile: A context manager sounds like a fantastic idea. Dang, that's clever!

tim-schilling (approved):
:rocket:

docs/querying_history.rst
ddabble: Alright :) I agree, though credits for the idea goes to https://github.com/jazzband/django-simple-history/issues/642#issuecomment-922110914 😄

Issue #1031, opened 2022-08-24
HistoricOneToOneField


valberg:
**Problem Statement**
We have just gotten `HistoricForeignKey` which is great! It would be nice to have a similar for `OneToOneField`.

**Describe the solution you'd like**
Implementation of `HistoricOneToOneField` which does the same as `HistoricForeignKey` just within the bounds of a `OneToOneField`.

quadracik:
No problem. We've been using it in production for over a year and I didn't experience any problems with it. Unfortunately I don't have enought free time to write propper tests and open a good pull request.

JordanHyatt:
Found it....you I think you needed `self.related.related_model`  instead of `self.related.model`

JordanHyatt:
See https://github.com/jazzband/django-simple-history/pull/1394

quadracik:
@JordanHyatt , actually now you mentioned it, I checked and the final solution we use also returns the queryset instead assigning it and uses `self.related.related_model`.
There is one more difference - the final code doesn't do `try .. except`. Instead, it only does what is in the `except` clause. I don't remember why, but it must have been causing some problems.. Or maybe it only applies to reverse FKs as they may have `prefetch_related` used on them and for 1to1 it never goes this way. Not sure now. But please have it in mind.
My final code for `HistoricReverseOneToOneDescriptor` below:
```
class HistoricReverseOneToOneDescriptor(ReverseOneToOneDescriptor):
    """
    Overrides get_queryset to provide historic query support, should the
    instance be historic (and therefore was generated by a timepoint query)
    and the other side of the relation also uses a history manager.
    """

    def get_queryset(self, **hints):
        instance = hints.get("instance")
        if instance:
            history = getattr(
                instance, SIMPLE_HISTORY_REVERSE_ATTR_NAME, None
            )
            histmgr = getattr(
                self.related.related_model,
                getattr(
                    self.related.related_model._meta,
                    "simple_history_manager_attribute",
                    "_notthere",
                ),
                None,
            )
            if history and histmgr:
                return histmgr.as_of(
                    getattr(history, "_as_of", history.history_date)
                )
        return super().get_queryset(**hints)
```

JordanHyatt:
That's funny, all 3 of those things I noticed and adjusted.  See the PR I submitted.  Thanks for posting, it definitely gave me a good headstart

quadracik:
Great, haven't noticed it. Thanks for your work!

ddabble:
[...]

PR #1265, opened 2023-09-28
add historic one to one field

david-homelend:
add historic one to one field

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -3,6 +3,7 @@ import importlib
 import uuid
 import warnings
 from functools import partial
+from typing import TypeVar
 
 from django.apps import apps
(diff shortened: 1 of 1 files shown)


ddabble (changes requested):
Without having taken the time to examine this, I can personally not confidently say that this code would work - especially considering how a previous suggestion trying to implement it was considerably longer (see https://github.com/jazzband/django-simple-history/issues/1031#issuecomment-1505459663) - so could you provide some tests for this? 🙂

JordanHyatt:
See https://github.com/jazzband/django-simple-history/pull/1394

tim-schilling:
Closing in favor of #1394

Issue #1181, opened 2023-05-29
Calling update_change_reason on a null JSONField with a postgres DB raise an AttributeError


helenecourau:
**Describe the bug**
I use a postgres database. I call the update_change_reason method in a post_create_historical_record signal and my model has a nullable JSONField. It raises an AttributeError 'NoneType' object has no attribute 'history_change_reason'. 

**To Reproduce**
1. Create a model with a null JSONField on a project with a postgres database.
2. Register the history on all fields.
3. Call the update_change_reason method.
4. It raises an AttributeError 'NoneType' object has no attribute 'history_change_reason'. 

**Expected behavior**
No error raised.

**Screenshots**

**Environment (please complete the following information):**
 - OS: MacOS 13.3.1
 - Django Simple History Version: 3.3.0
 - Django Version: 4.0.1
 - Database Version: PostgreSQL 14.8

**Additional context**

The problem is that update_change_reason makes this request : model.objects.filter(json_field=None)
For a postgres JSONField, None means that there is a null value and in this case there is no null value just an empty field.

I tried to override locally the update_change_reason method to make this request for the null jsonfield: model.objects.filter(json_field__isnull=True) and it works.
If it's usefull I can open a pull request with this change.

PR #1282, opened 2023-12-02
Attempts to fix update_change_reason for nullable JSONField fields (issue #1181)

martinschwinzerl:
## Description
As reported in issue #1181, the implementation of the utility `update_change_reason()` method does not work for models with a nullable JSONField. This is caused by the slightly more convoluted querying mechanisms that has to be applied for JSONFields, please cf. here: [django dev documentation](https://docs.djangoproject.com/en/dev/topics/db/queries/#querying-jsonfield)

This attempts to fix this by checking 
- if a field is a JSONField
- is nullable
- and then using different query language, depending on the `null` status of the field in question.

The strategy assumes that
- a nullable JSONField with a `value` of `None` is not set at all
- a non-nullable JSONField with a `value` of `None` is actually a "NULL" dictionary

**Note**: I am aware that this is not the only way to handle this. But it is the convention I am using in my project and this fixed the issue for me. It may be worthwhile to implement a policy switch to decide how to interpret `None` in this specific case.

## Related Issue
- This attempts to fix issue #1181

## Motivation and Context
I did run into the same issue as the original reporter of the bug report and was stuck. The fix, while admittedly maybe not yet fully universal, depending on the interpretation of a "nullable" JSONField, did solve the issue for me

## How Has This Been Tested?
Tested using my local installation. No unit test implemented but this can be done, if required

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [] I have run the `pre-commit run` command to format and lint.

### DIFF
--- a/simple_history/utils.py
+++ b/simple_history/utils.py
@@ -1,5 +1,5 @@
 from django.db import transaction
-from django.db.models import Case, ForeignKey, ManyToManyField, Q, When
+from django.db.models import Case, ForeignKey, ManyToManyField, Q, When, JSONField, Value
 from django.forms.models import model_to_dict
 
 from simple_history.exceptions import AlternativeManagerError, NotHistoricalModelError
@@ -18,6 +18,11 @@ def update_change_reason(instance, reason):
         if field.primary_key is True:
             if value is not None:
                 attrs[field.attname] = value
+        elif isinstance(field, JSONField):
+            if value is None and field.null is True:
+                attrs[f"{field.attname}__isnull"] = True
+            else:
+                attrs[field.attname] = Value(value, JSONField())
         else:
             attrs[field.attname] = value
 


martinschwinzerl:
Sorry, I had some issues with running the pre-commit hooks on my development machine as I do not have a python 3.8 environment available. 
```
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/home/martin/.virtualenvs/ps_sps_users/bin/python', '-mvirtualenv', '/home/martin/.cache/pre-commit/repo5rnoh46q/py_env-python3.8', '-p', 'python3.8')
return code: 1
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.8'
stderr: (none)
Check the log at /home/martin/.cache/pre-commit/pre-commit.log
```

I did locally modify the commit hooks to use python 3.11 and everything seemed to work but now my local branch contains this change to the pre-commit hooks and everything is a bit of a mess. 

I am in a hurry but may have time to run again later next week on an older machine with python 3.8, any suggestions or pointers on how to handle this in a less awkward fashion would be highly appreciated!

PR #1319, opened 2024-03-25
Improve performance of latest_of_each for PostgreSQL

Flauschbaellchen:
When getting historic PKs, do not evaluate the query but let the database decide how to improve this sub-query.
This improves the performance as the roundtrip into Python lists are skipped.

## Motivation and Context
For a huge dataset, this would be a memory and performance issue when calling `values_list` instead of letting the database handle the sub-query directly.

## How Has This Been Tested?
No test changes needed.

## Checklist:

- [x] My change requires a change to the documentation. (not needed)
- [x] I have updated the documentation accordingly. (not needed)
- [x] I have added tests to cover my changes. (no changes)
- [x] I have added my name and/or github handle to `AUTHORS.rst` (I think that's too small of a change... ;))
- [ ] I have added my change to `CHANGES.rst` (idk if this is too small?)

### DIFF
--- a/simple_history/manager.py
+++ b/simple_history/manager.py
@@ -77,7 +77,7 @@ class HistoricalQuerySet(QuerySet):
             latest_pk_attr_historic_ids = (
                 self.order_by(self._pk_attr, "-history_date", "-pk")
                 .distinct(self._pk_attr)
-                .values_list("pk", flat=True)
+                .values("pk")
             )
             latest_historics = self.filter(history_id__in=latest_pk_attr_historic_ids)
         else:


Flauschbaellchen:
Hi @ddabble, might I kindly bump this PR? As it is only a very small change, it should be quick to review. :bow: 
If I missed sth. I would be very happy to improve.

ddabble:
Hey, thanks for the PR 😊

> When getting historic PKs, do not evaluate the query but let the database decide how to improve this sub-query.
> This improves the performance as the roundtrip into Python lists are skipped.

Are you sure this is an accurate description of the existing behavior and the behavior of your changes..? Because, as far as I know, both `values_list()` and `values()` return querysets, which, when passed directly to `filter()`, are not evaluated (and also not converted into Python lists). Furthermore, looking at Django's code, it seems like [this section](https://github.com/django/django/blob/5.0.4/django/db/models/query.py#L1365-L1390) (or [the same section for 3.2](https://github.com/django/django/blob/3.2/django/db/models/query.py#L845-L865)) is the main difference between `values_list()` and `values()` (apart from the iterator assigned to `_iterable_class`, which shouldn't be used at all when passing the queryset directly to `filter()`, as mentioned above), and no parts of the marked code are affected by the size of the queryset, as far as I can tell.

So, could you provide some more concrete details on how this improves the performance? 🙂

Flauschbaellchen:
Hi @ddabble,

thanks for taking your time to look at the PR.
I'm truly sorry - you are totally right! I somehow missed this.
It seems that I had some other side effects in my project while testing this out and was too fast to open up a PR instead of running it once more in a cleaner environment.
By reading the code you referenced and also running the queries manually its obvious that the filters/sub-queries are evaluated lazely. :man_facepalming:  
Thanks for pointing it out :bow: 

I'm closing this PR as it does not make any sense :sweat_smile:

ddabble:
Alright, no worries! Thanks for taking the time to contribute to the project, and please don't hesitate to do so in the future 😊

PR #736, opened 2020-11-06
Add ignore_saving_historical_record_on_delete flag

pvanderw:
## Description
Add ignore_saving_historical_record_on_delete flag on HistoricalRecord model which ignores saving history when a model is deleted. Fixes performance issue when cascading deletions create a large number of history instances.

## Related Issue
Similar to issue #717 

## Motivation and Context
Deleting a model that causes a lot of cascading deletions is slow since it creates a new history instance for each delete, potentially causing thousands of database queries.

## How Has This Been Tested?
Added new unit tests with the flag set to both True/False

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have run the `make format` command to format my code
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [ ] I have added my name and/or github handle to `AUTHORS.rst`
- [x] I have added my change to `CHANGES.rst`
- [ ] All new and existing tests passed.

PR #1387, opened 2024-09-10
`disable_history()` context manager

### LINKED ISSUE #642: Best way to dynamically disable creation of history records?

### LINKED ISSUE #993: Skipping history creation when deleting an object
**Problem Statement**
There is no attribute to skip saving a delete history entry when deleting an object.

Either `skip_history_when_saving` on the object should be honoured (like in the `post_save` method) or a new `skip_history_when_deleting` attribute should be introduced. 

I'm hacking the settings to temporarily set `SIMPLE_HISTORY_ENABLED = False` to avoid saving it.

**Additional context**
Implementation should be pretty straightforward like in the `post_save` method: https://github.com/jazzband/django-simple-history/blob/master/simple_history/models.py#L541

### LINKED ISSUE #1365: Add support for skip_history_when_saving in M2M fields
**Problem Statement**
**skip_history_when_saving** works great for regular fields in history, but I would like the ability to update m2m fields without saving history.

Add support for m2m fields during add/remove with a flag similar to **skip_history_when_saving**

Only approach right now is to not track m2m fields/track them separately in order to avoid unneeded history for my use case.

ddabble:
## Description

Added a `disable_history()` context manager, which disables history record creation while it's active; see usage in the docs under "Disable Creating Historical Records" (8ea956c4067f166642336a32c2c478a55dc14c56).

Also:

- Added `delete_without_historical_record()` to all history-tracked model objects, which complements `save_without_historical_record()` (e2894b54009298c924c5e70656492a085c404529)

**Breaking changes:**

- Removed `HistoryManager.get_super_queryset()` (23c37dda6c1d70d934ac53bb2f1ef4469a890f9e)
- Renamed the `utils` functions `get_history_manager_from_history()` to `get_historical_records_of_instance()` and `get_app_model_primary_key_name()` to `get_pk_name()` (23c37dda6c1d70d934ac53bb2f1ef4469a890f9e)

**Deprecations:**

- Deprecated the undocumented `HistoricalRecords.thread` - use `HistoricalRecords.context` instead. The former attribute will be removed in version 3.10 (74a2e381bf78b0ca6afe7bb0f12019cdcea72491)
- Deprecated `skip_history_when_saving` in favor of the newly added `disable_history()` context manager. The former attribute will be removed in version 4.0 (8ea956c4067f166642336a32c2c478a55dc14c56)

**Fixes and improvements:**

- Moved the "Save without creating historical records" subsection of "Querying History" in the docs to a new section: "Disable Creating Historical Records" (b36a280a4c0eea9a467bb4ee66bc40b9c028edae)
- The `utils` functions `get_history_manager_for_model()` and `get_history_model_for_model()` now explicitly support being passed model instances instead of just model types (78286f6f5b691d932c89e56e3e7f5dbbc5608047)

**Other changes:**
- Documented overriding `create_historical_record()` (7755c1e496a4dfeb2e60f3da8546123552048631)

## Related Issue

* Closes #642
* Closes #993
* Closes #1365
* Closes #736

## Motivation and Context

Having an easier and more universal way of disabling the creation of historical records in various contexts.

## How Has This Been Tested?

See the added tests.

## Types of changes

## Checklist:

### DIFF
--- a/docs/disabling_history.rst
+++ b/docs/disabling_history.rst
@@ -8,34 +8,40 @@ These methods are automatically added to a model when registering it for history
 (i.e. defining a ``HistoricalRecords``  manager on the model),
 and can be called instead of ``save()`` and ``delete()``, respectively.
 
-Setting the ``skip_history_when_saving`` attribute
---------------------------------------------------
+Using the ``disable_history()`` context manager
+-----------------------------------------------
 
-If you want to save or delete model objects without triggering the creation of any
-historical records, you can do the following:
+``disable_history()`` has three ways of being called:
 
-.. code-block:: python
+#. With no arguments: This will disable all historical record creation
+   (as if the ``SIMPLE_HISTORY_ENABLED`` setting was set to ``False``; see below)
+   within the context manager's ``with`` block.
+#. With ``only_for_model``: Only disable history creation for the provided model type.
+#. With ``instance_predicate``: Only disable history creation for model instances passing
+   this predicate.
 
-    poll.skip_history_when_saving = True
-    # It applies both when saving...
-    poll.save()
-    # ...and when deleting
-    poll.delete()
-    # We recommend deleting the attribute afterward
-    del poll.skip_history_when_saving
-
-This also works when creating an object, but only when calling ``save()``:
+See some examples below:
 
 .. code-block:: python
 
-    # Note that `Poll.objects.create()` is not called
-    poll = Poll(question="Why?")
-    poll.skip_history_when_saving = True
-    poll.save()
-    del poll.skip_history_when_saving
+    from simple_history.utils import disable_history
 
-.. note::
-    Historical records will always be created when calling the ``create()`` manager method.
+    # No historical records are created
+    with disable_history():
+        User.objects.create(...)
+        Poll.objects.create(...)
+
+    # A historical record is only created for the poll
+    with disable_history(only_for_model=User):
+        User.objects.create(...)
+        Poll.objects.create(...)
+
+    # A historical record is created for the second poll, but not for the first poll
+    # (remember to check the instance type in the passed function if you expect
+    # historical records of more than one model to be created inside the `with` block)
+    with disable_history(instance_predicate=lambda poll: "ignore" in poll.question):
+        Poll.objects.create(question="ignore this")
+        Poll.objects.create(question="what's up?")
 
 The ``SIMPLE_HISTORY_ENABLED`` setting
 --------------------------------------
--- a/docs/disabling_history.rst
+++ b/docs/disabling_history.rst
@@ -1,0 +1,51 @@
+Disable Creating Historical Records
+===================================
+
+Save without creating historical records
+----------------------------------------
+
+If you want to save model objects without triggering the creation of any historical
+records, you can do the following:
+
+.. code-block:: python
+
+    poll.skip_history_when_saving = True
+    poll.save()
+    # We recommend deleting the attribute afterward
+    del poll.skip_history_when_saving
+
+This also works when creating an object, but only when calling ``save()``:
+
+.. code-block:: python
+
+    # Note that `Poll.objects.create()` is not called
+    poll = Poll(question="Why?")
+    poll.skip_history_when_saving = True
+    poll.save()
+    del poll.skip_history_when_saving
+
+.. note::
+    Historical records will always be created when calling the ``create()`` manager method.
+
+Alternatively, call the ``save_without_historical_record()`` method on each object
+instead of ``save()``.
+This method is automatically added to a model when registering it for history-tracking
+(i.e. defining a ``HistoricalRecords``  manager field on the model),
+and it looks like this:
+
+.. code-block:: python
+
+    def save_without_historical_record(self, *args, **kwargs):
+        self.skip_history_when_saving = True
+        try:
+            ret = self.save(*args, **kwargs)
+        finally:
+            del self.skip_history_when_saving
+        return ret
+
+Or disable the creation of historical records for *all* models
+by adding the following line to your settings:
+
+.. code-block:: python
+
+    SIMPLE_HISTORY_ENABLED = False
--- a/docs/disabling_history.rst
+++ b/docs/disabling_history.rst
@@ -1,16 +1,26 @@
 Disable Creating Historical Records
 ===================================
 
-Save without creating historical records
-----------------------------------------
+``save_without_historical_record()`` and ``delete_without_historical_record()``
+-------------------------------------------------------------------------------
 
-If you want to save model objects without triggering the creation of any historical
-records, you can do the following:
+These methods are automatically added to a model when registering it for history-tracking
+(i.e. defining a ``HistoricalRecords``  manager on the model),
+and can be called instead of ``save()`` and ``delete()``, respectively.
+
+Setting the ``skip_history_when_saving`` attribute
+--------------------------------------------------
+
+If you want to save or delete model objects without triggering the creation of any
+historical records, you can do the following:
 
 .. code-block:: python
 
     poll.skip_history_when_saving = True
+    # It applies both when saving...
     poll.save()
+    # ...and when deleting
+    poll.delete()
     # We recommend deleting the attribute afterward
     del poll.skip_history_when_saving
 
@@ -27,23 +37,10 @@ This also works when creating an object, but only when calling ``save()``:
 .. note::
     Historical records will always be created when calling the ``create()`` manager method.
 
-Alternatively, call the ``save_without_historical_record()`` method on each object
-instead of ``save()``.
-This method is automatically added to a model when registering it for history-tracking
-(i.e. defining a ``HistoricalRecords``  manager field on the model),
-and it looks like this:
+The ``SIMPLE_HISTORY_ENABLED`` setting
+--------------------------------------
 
-.. code-block:: python
-
-    def save_without_historical_record(self, *args, **kwargs):
-        self.skip_history_when_saving = True
-        try:
-            ret = self.save(*args, **kwargs)
-        finally:
-            del self.skip_history_when_saving
-        return ret
-
-Or disable the creation of historical records for *all* models
+Disable the creation of historical records for *all* models
 by adding the following line to your settings:
 
 .. code-block:: python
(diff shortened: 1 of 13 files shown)


tim-schilling:
@ddabble I'm pretty swamped through the first week of October right now if you can wait until then.

ddabble:
Sure, no rush 😊

docs/disabling_history.rst
```diff
@@ -0,0 +1,77 @@
...
+-----------------------------------------------
+
+``disable_history()`` has three ways of being called:
+
+#. With no arguments: This will disable all historical record creation
+   (as if the ``SIMPLE_HISTORY_ENABLED`` setting was set to ``False``; see below)
+   within the context manager's ``with`` block.
+#. With ``only_for_model``: Only disable history creation for the provided model type.
+#. With ``instance_predicate``: Only disable history creation for model instances passing
+   this predicate.
```
tim-schilling, line 21: This argument name doesn't seem very beginner friendly. The best thing I can come up with is `should_disable_callback`. Or maybe `should_disable`?

We should also document what arguments this function must take.

docs/disabling_history.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Using the ``disable_history()`` context manager
+-----------------------------------------------
+
+``disable_history()`` has three ways of being called:
+
+#. With no arguments: This will disable all historical record creation
+   (as if the ``SIMPLE_HISTORY_ENABLED`` setting was set to ``False``; see below)
+   within the context manager's ``with`` block.
+#. With ``only_for_model``: Only disable history creation for the provided model type.
```
tim-schilling, line 19: Perhaps this is better as `models` to match the signature of `isinstance`? This would allow people to disable multiple models in a single context manager and it shouldn't be too much additional complexity on our end.

docs/disabling_history.rst
```diff
@@ -0,0 +1,77 @@
...
+
+    class Poll(models.Model):
+        # ...
+        history = CustomHistoricalRecords()
+
+The ``SIMPLE_HISTORY_ENABLED`` setting
+--------------------------------------
+
+Disable the creation of historical records for *all* models
+by adding the following line to your settings:
```
tim-schilling, line 73: Do we want to indicate this defaults to `True`?

tim-schilling: Oh, this can be ignored. I realized that it's just moving.

simple_history/manager.py
```diff
@@ -118,16 +114,13 @@ def __init__(self, model, instance=None):
...
-
     def get_queryset(self):
-        qs = self.get_super_queryset()
+        qs = super().get_queryset()
         if self.instance is None:
             return qs
 
-        key_name = get_app_model_primary_key_name(self.instance)
-        return self.get_super_queryset().filter(**{key_name: self.instance.pk})
+        pk_name = utils.get_pk_name(self.instance._meta.model)
```
tim-schilling, line 122:
```suggestion
        pk_name = utils.get_pk_name(self.model)
```
Does that work just as well?

simple_history/manager.py
```diff
@@ -222,15 +215,19 @@ def bulk_history_create(
         If called by bulk_update_with_history, use the update boolean and
         save the history_type accordingly.
         """
-        if not getattr(settings, "SIMPLE_HISTORY_ENABLED", True):
-            return
+        info = utils.DisableHistoryInfo.get()
+        if info.disabled_globally:
+            return []
```
tim-schilling, line 220: This is a breaking API change and should be documented in the release notes.

simple_history/models.py
```diff
@@ -4,9 +4,21 @@
...
+    Dict,
+    Iterable,
+    List,
+    Sequence,
+    Type,
+    Union,
+)
 
 import django
+from asgiref.local import Local
```
tim-schilling, line 21: You don't have to change to it, but why not a [`ContextVar`](https://docs.python.org/3/library/contextvars.html)?

simple_history/models.py
```diff
@@ -178,20 +196,34 @@ def contribute_to_class(self, cls, name):
...
 
             Make sure you know what you're doing before using this method.
             """
-            self.skip_history_when_saving = True
-            try:
-                ret = self.save(*args, **kwargs)
-            finally:
-                del self.skip_history_when_saving
-            return ret
+            with utils.disable_history(instance_predicate=get_instance_predicate(self)):
```
tim-schilling, line 213: If the instance being saved has cascading receivers configured on `post_save`, those would likely generate history here while the previous version they wouldn't.

I suspect this manager should disable all history rather than trying to only disable it for this instance.

simple_history/models.py
```diff
@@ -178,20 +196,34 @@ def contribute_to_class(self, cls, name):
...
 
-        setattr(cls, "save_without_historical_record", save_without_historical_record)
+        def delete_without_historical_record(self: models.Model, *args, **kwargs):
+            """
+            Delete the model instance without creating a historical record.
+
+            Make sure you know what you're doing before using this method.
+            """
+            with utils.disable_history(instance_predicate=get_instance_predicate(self)):
+                return self.delete(*args, **kwargs)
```
tim-schilling, line 223: Same as above.

simple_history/tests/tests/test_deprecation.py
```diff
@@ -1,13 +1,55 @@
...
+        with self.assertWarns(DeprecationWarning):
+            context = HistoricalRecords().thread
+            self.assertIs(context, HistoricalRecords.context)
+        with self.assertWarns(DeprecationWarning):
+            context = getattr(HistoricalRecords(), "thread")
+            self.assertIs(context, HistoricalRecords.context)
+
+        # DEV: `_DeprecatedThreadDescriptor` and the `thread` attribute of
+        #      `HistoricalRecords` should be removed when 3.10 is released
+        self.assertLess(__version__, "3.10")
```
tim-schilling, line 35: Nice, I like this.

simple_history/tests/tests/utils.py
```diff
@@ -26,13 +28,32 @@ def assertRecordValues(self, record, klass: Type[Model], values_dict: dict):
...
+            else:
+                self.assertEqual(value, expected_value)
+
+        history_object = record.history_object
+        self.assertEqual(history_object.__class__, klass)
+        for field_name, expected_value in values_dict_copy.items():
+            if field_name not in ("history_type", "history_change_reason"):
+                self.assertEqual(getattr(history_object, field_name), expected_value)
+
+    def _assert_m2m_record(self, record, field_name: str, expected_value: List[Model]):
```
tim-schilling, line 49: might this be better named as `_assert_m2m_relation`? Or `_assert_m2m_queryset`? I don't mean to bikeshed, but as it currently stands, it makes it sounds like it's comparing a single value. However, it's really fetching all the related objects to compare. Having the name imply that it'll do a DB hit will be useful for the future.

simple_history/utils.py
```diff
@@ -1,11 +1,184 @@
...
+    instance_predicate: Callable[[Model], bool] = None,
+) -> Iterator[None]:
+    """
+    Disable creating historical records while this context manager is active.
+
+    Note: ``only_for_model`` and ``instance_predicate`` cannot both be provided.
+
+    :param only_for_model: Only disable history creation for this model type.
+    :param instance_predicate: Only disable history creation for model instances passing
+        this predicate.
```
tim-schilling, line 32: Similar to our documentation, we should include the expected signature here.

simple_history/utils.py
```diff
@@ -1,11 +1,184 @@
...
+        _StoredDisableHistoryInfo.get() is None
+    ), "Nesting 'disable_history()' contexts is undefined behavior"
+
+    if only_for_model:
+        # Raise an error if it's not a history-tracked model
+        get_history_manager_for_model(only_for_model)
+        if instance_predicate:
+            raise ValueError(
+                "'only_for_model' and 'instance_predicate' cannot both be provided"
+            )
```
tim-schilling, line 44: I'm curious why you decided on this. I'm generally fine with it, but we could `AND` them together.

simple_history/utils.py
```diff
@@ -1,11 +1,184 @@
...
+        return HistoricalRecords.context
+
+
+@dataclass(
+    frozen=True,
+    # DEV: Replace this with just `kw_only=True` when the minimum required
+    #      Python version is 3.10
+    **({"kw_only": True} if sys.version_info >= (3, 10) else {}),
+)
+class DisableHistoryInfo:
```
tim-schilling, line 97: I didn't have an issue with this before, but looking at the first method of `not_disabled` and I'm not sure this is how things should be done. My concern is that this control is solely focused on history being disabled. However, that also means it's focused on history being enabled.

So maybe a name like `HistoryControl`, `HistoryState` or `HistoryInstrumentation`? Then we can change `not_disabled` to `enabled` :grin:

simple_history/utils.py
```diff
@@ -1,11 +1,184 @@
...
+    def not_disabled(self) -> bool:
+        """
+        A value of ``True`` means that historical record creation is not disabled
+        in any way.
+        If ``False``, check ``disabled_globally`` and ``disabled_for()``.
+        """
+        return not self._disabled_globally and not self._instance_predicate
+
+    @property
+    def disabled_globally(self) -> bool:
```
tim-schilling, line 118: nitpick: There is a bit of a precedence set in the API for a `is_` prefix on these types of boolean property checks.

tim-schilling (changes requested):
I feel somewhat strongly about the API naming for the `disable_history()` arguments as well as the `DisableHistoryInfo` class naming. Most of everything else is suggestions or questions for clarification.

Great work as always :muscle:. It's always fun and educational to review your code!

tim-schilling:
@kut no, I think this is something we want to merge in. I don't think ddabble nor I have a ton of time to dedicate to feature development here though. I can continue to help on code reviews though.

PR #1402, opened 2024-10-04
Fixes #527 history button not working

### LINKED ISSUE #527: Admin page History button has broken link for historical objects.
**Bug**
When viewing a historical object in the admin, the "HISTORY" button on the top right does nothing on-click. 

**To Reproduce**
1. Go to admin page.
2. Click on any model.
3. Click on an instance of the model to get to the admin form.
4. Click the "HISTORY" button in the top right.
5. Click on some past instance.
6. Click the "HISTORY" button in the top right again.
7. Nothing happens.

**Expected behavior**
On clicking the button, you return to the list of historical objects.

**Screenshots**
HTML from Chrome developer, showing an empty href:

image

 - OS: Windows 10
 - Browser (if applicable): Chrome
 - Django Simple History Version: 2.7.0
 - Django Version: 2.1.5
 - Database Version: PostgreSQL 10.4

**Additional context**
I'm not 100% sure if this is coming from your side or mine. Every other piece of history-functionality appears to be working - I cant see what would cause this bug.

trumpet2012:
## Description
This PR fixes the history button that gets shown when viewing a historical version of a model so that it correctly links back to the history listing for that object.

## Related Issue
#527 

## Motivation and Context
Fixes a broken link that users expect to work.

## How Has This Been Tested?
I tested this by installing it as a development dependency in our django project. I also ran `make tests` however only the sqlite versions worked since I haven't setup postgres/msql locally.

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/templates/simple_history/object_history_form.html
+++ b/simple_history/templates/simple_history/object_history_form.html
@@ -1,5 +1,5 @@
 {% extends "admin/change_form.html" %}
-{% load i18n %}
+{% load i18n admin_urls %}
 {% load url from simple_history_compat %}
 
 {% block breadcrumbs %}
@@ -19,6 +19,14 @@
 
 {% block submit_buttons_bottom %}
   {% include "simple_history/submit_line.html" %}
+{% endblock %}
+
+{% block object-tools-items %}
+{# We override this block from the django template to fix up the history link #}
+<li>
+    <a href="{% add_preserved_filters history_url %}" class="historylink">{% translate "History" %}</a>
+</li>
+{% if has_absolute_url %}<li><a href="{{ absolute_url }}" class="viewsitelink">{% translate "View on site" %}</a></li>{% endif %}
 {% endblock %}
 
 {% block form_top %}
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -146,6 +146,7 @@ Authors
 - `DanialErfanian <https://github.com/DanialErfanian>`_
 - `Sridhar Marella <https://github.com/sridhar562345>`_
 - `Mattia Fantoni <https://github.com/MattFanto>`_
+- `Trent Holliday <https://github.com/trumpet2012>`_
 
 Background
 ==========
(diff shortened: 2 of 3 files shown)


simple_history/templates/simple_history/object_history_form.html
```diff
@@ -21,6 +21,14 @@
   {% include "simple_history/submit_line.html" %}
 {% endblock %}
 
+{% block object-tools-items %}
+{# We override this block from the django template to fix up the history link #}
+<li>
+    <a href="{% add_preserved_filters history_url %}" class="historylink">{% translate "History" %}</a>
+</li>
+{% if has_absolute_url %}<li><a href="{{ absolute_url }}" class="viewsitelink">{% translate "View on site" %}</a></li>{% endif %}
```
tim-schilling, line 29: Is there a reason to not use `{% change_form_object_tools %}` from `admin_notify`?

trumpet2012: I'm not sure what you mean by `admin_notify` I didn't see a django app by that name. 

The `{% change_form_object_tools %}` template tag is the default behavior of the `object-tools-items` block. We can't use it because that template is always overriding the `history_url` context variable to point to the django history version instead of using the simple history version from the context. https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/change_form_object_tools.html#L4

It's definitely not ideal to have to implement all of the extra pieces that would normally be handled by that template tag.

tim-schilling: Ah, my bad. I didn't go deep enough.

tim-schilling (approved):
This looks good! Thank you @trumpet2012 for your patience

tim-schilling:
I found another way to do this.

We could define a template at `templates/admin/change_form_object_tools.html` and if the `simple_history` app is installed before the admin, it allows us to override the template without having to override that block.

However, I think the additional installation requirements are a bit too burdensome.

PR #1434, opened 2025-01-02
Fixed reason field being shared between models.

### LINKED ISSUE #1433: Bug using django ORM to update inherited history_change_reason field
**Describe the bug**
When using `HistoricalRecords(inherit=True, history_change_reason_field=...)`, calling `update(history_change_reason=...)` on the history model will update an incorrect model due to the reason field being shared between any derived model.

**To Reproduce**
With the models
``` python
from django.db import models
from simple_history.models import HistoricalRecords


class Base(models.Model):
    history = HistoricalRecords(
        inherit=True, history_change_reason_field=models.TextField()
    )


class Derived1(Base):
    pass


class Derived2(Base):
    pass
```

Run the following
``` python
d1 = Derived1.history.create(history_change_reason='reason1', id=1, history_date=timezone.now())
d2 = Derived2.history.create(history_change_reason='reason2', id=1, history_date=timezone.now())

Derived1.history.update(history_change_reason='new_reason1')

d1.refresh_from_db()
d2.refresh_from_db()

print(d1.history_change_reason)
print(d2.history_change_reason)
```

Resulting in
```
reason1
new_reason1
```

**Expected behavior**
The correct historical model should be updated.

 - Django Simple History Version: 3.7.0
 - Django Version: 5.1.4
 - Database Version: SQLite 3.47.2

noamkush:
## Description

This fixes the custom reason field being shared between inherited models. It causes issues when calling `update()` on the historical model. When generating the field, a `clone()` call is added so the field wouldn't be shared.

## Related Issue

Fixes #1433 

## How Has This Been Tested?

Added unit tests reproducing the issue. The tests fail on master branch.

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -411,7 +411,7 @@ class HistoricalRecords:
     def _get_history_change_reason_field(self):
         if self.history_change_reason_field:
             # User specific field from init
-            history_change_reason_field = self.history_change_reason_field
+            history_change_reason_field = self.history_change_reason_field.clone()
         elif getattr(
             settings, "SIMPLE_HISTORY_HISTORY_CHANGE_REASON_USE_TEXT_FIELD", False
         ):
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -111,6 +111,7 @@ Authors
 - Nathan Villagaray-Carski (`ncvc <https://github.com/ncvc>`_)
 - Nianpeng Li
 - Nick Träger
+- Noam Kushinsky (`noamkush <https://github.com/noamkush>`_)
 - Noel James (`NoelJames <https://github.com/NoelJames>`_)
 - Ofek Lev (`ofek <https://github.com/ofek>`_)
 - Phillip Marshall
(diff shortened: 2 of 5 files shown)


simple_history/models.py
```diff
@@ -411,7 +411,7 @@ def copy_fields(self, model):
     def _get_history_change_reason_field(self):
         if self.history_change_reason_field:
             # User specific field from init
```
tim-schilling, line 413:
```suggestion
            # User specific field from init.
            # It's cloned so a reason from an inherited model isn't shared.
```
Does that match the fix?

tim-schilling (commented):
This looks good to me. One question on whether a comment may be useful.

tim-schilling:
I committed a the conflict resolution to allow me to trigger the workflows to run to confirm the tests pass.

simple_history/models.py
noamk-hl: Yeah, sure

noamkush: Yeah, sure

Issue #1475, opened 2025-04-11
`update_change_reason`: fails to fetch latest history entry for nullable fields


Senor-Ducky:
**Describe the bug**
The `update_change_reason` method in `utils.py` fails to update change reason in case any field is nullable in a model or is set to default to `None`. This could be due to the way django handles null values, as direct comparison to `None` is not allowed when filtering.

**To Reproduce**
1. Create a model with any nullable field or a field that defaults to `None` and enable historical records.
2. Try to use the `update_change_reason` on an instance of the model.

**Expected behavior**
The `update_change_reason` should handle models with nullable fields/fields with default set to None gracefully by applying the correct filter query for fetching historical records.

**Screenshots**
![Image](https://github.com/user-attachments/assets/c3dc0506-63d1-4659-b7e8-61fbc8f863b9)

**Environment (please complete the following information):**
 - OS: Win 11 24H2
 - Django Simple History Version: 3.8.0
 - Django Version: 5.0.11
 - Database Version: PostgreSQL 16.8

**Additional context**
In the provided screenshot you can see, the `History Attrs` logged from within the `update_change_reason` method indicating that it is using the wrong filter query for filtering/matching null value, instead of `field=None` it should have been `field__isnull`=True, which prevented the historical record from showing up causing an exception.

Senor-Ducky:
> > This could be due to the way django handles null values, as direct comparison to `None` is not allowed when filtering.
> > [...]
> > indicating that it is using the wrong filter query for filtering/matching null value, instead of `field=None` it should have been `field__isnull`=True
> 
> ?? Django handles `filter(field=None)` in exactly the same way as `filter(field__isnull=True)` (see [the queryset docs](https://docs.djangoproject.com/en/stable/ref/models/querysets/#exact)).
> 
> Also, [the test you added in your PR](https://github.com/jazzband/django-simple-history/pull/1476/commits/aa5aa82c55cb9824f843adc02f7744ae583582b5#diff-df4755711a972ea6fb69be8f02dcbc614934c4ce8ea9b9902308570a442a71caR644-R652) passes without the other changes in the PR 😅
> 
> Now, looking at the screenshot you posted, it seems like the actual issue is that the instance you tried to update `history_change_reason` for didn't have any historical records. Could you try verifying whether that's the case? 🙂

Yes this is the case, it didn't have any historical records previously.

ddabble:
[...]

PR #1476, opened 2025-04-11
Fix `update_change_reason`: fails to fetch latest history entry for nullable fields

Senor-Ducky:
## Description
The `update_change_reason` method in `utils.py` fails to update change reason in case any field is nullable in a model or is set to default to `None`. This could be due to the way django handles null values, as direct comparison to None is not allowed when filtering.

## Related Issue
#1475 

## Motivation and Context
Encountered this issue when dealing with a model with nullable fields in one of my projects

## How Has This Been Tested?
Has been tested both within my project environment via test cases simulating this scenario, also tested within the suite by adding a new similar test case.

image

## Types of changes

## Checklist:


ddabble:
Closing - see the issue discussion.

Issue #1532, opened 2025-10-02
UUIDv7 for `history_id`


epiben:
**Problem Statement**
Currently, only UUIDv4 is supported, which is inferior to UUIDv7 w.r.t. performance because they're sortable.

**Describe the solution you'd like**
To enable the use of UUIDv7 for `history_id` fields. I'm happy to (and already have, in a fork) implement this feature. To not break existing code, I would add a new setting called `SIMPLE_HISTORY_HISTORY_ID_UUID_VERSION`, which must be only `4` or `7` and will be `4` by default.

**Describe alternatives you've considered**
Stick with the current UIDv4-only approach.

**Additional context**
N/A.

epiben:
Did anyone have the opportunity to look at this? And, mainly, the accompanying PR that adds this functionality?

chidoziemanagwu:
Hi @epiben , I'm an AI Product Leader and Senior Engineer. I’m a big advocate for moving to time-ordered UUIDs (v7) to solve the database fragmentation issues common with v4.

I see you’ve already started a fork, I’d love to help bring this over the finish line, specifically focusing on the setting logic (SIMPLE_HISTORY_HISTORY_ID_UUID_VERSION) and ensuring backwards compatibility. This is a great addition for high-performance auditing.

ddabble:
Closing for the reasons mentioned in https://github.com/django-commons/django-simple-history/pull/1545#issuecomment-4754351632.

Let's watch how the discussion potentially evolves here 🙂

PR #1545, opened 2025-11-03
Enable UUIDv7

epiben:
## Description
With this PR, users can choose between UUIDv4 and UUIDv7 for their PKs.

## Related Issue
This PR implements the request made in #1532.

## Motivation and Context
Using UUIDv7 yields faster I/O because they're sortable and, thus, friendlier to database indices. Being sortable, it's also possible to show the chronological order of historical records when these are created with below-minute resolution (the lowest allowed by `history_date`) or when multiple records are set to have the same `history_date` but with a meaningful order of insertion.

The implementation is non-breaking in that it adds a new settings for specifying the UUID version, falling back to UUIDv4 (the currently only option) if unspecified.

## How Has This Been Tested?
With the full test suite. The MySQL tests couldn't even run on my machine, so I'd be happy for some help to make that work locally if required.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -19,6 +19,7 @@ Authors
 - Anton Kulikov (`bigtimecriminal <https://github.com/bigtimecriminal>`_)
 - Ben Lawson (`blawson <https://github.com/blawson>`_)
 - Benjamin Mampaey (`bmampaey <https://github.com/bmampaey>`_)
+- Benjamin Skov Kaas-Hansen (`epiben <https://github.com/epiben>`_)
 - Berke Agababaoglu (`bagababaoglu <https://github.com/bagababaoglu>`_)
 - Bheesham Persaud (`bheesham <https://github.com/bheesham>`_)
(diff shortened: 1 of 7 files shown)


tim-schilling (commented):
The implementation looks generally fine.

ddabble:
@epiben I do see the advantages of supporting UUIDv7, but it seems unnecessary to add specific configuration for that one version (there are many other versions of UUID that have other advantages as well, as I'm sure you're aware).

We already support a `history_id_field` argument, where users can provide the exact PK field that they want, which is much more extensible. For users who would like a global default, it's possible to define a custom `HistoricalRecords` subclass and override the stuff they want - like @tim-schilling mentioned above.

Thank you for the PR anyway! 😊

---

I'm leaning toward closing this PR - unless you were thinking positively of it, @tim-schilling..?

tim-schilling:
@ddabble I'm good with closing it. We can always reopen it too

PR #1590, opened 2026-03-17
fix: catch correct DoesNotExist in HistoricReverseOneToOneDescriptor.…

awais786:
Closes #1588

Django's ReverseOneToOneDescriptor.__get__ catches self.related.related_model.DoesNotExist, but when HistoricDescriptorMixin.get_queryset() returns a queryset over the historical model (e.g. HistoricalB instead of B), only the historical model's DoesNotExist matches. Override __get__ in HistoricReverseOneToOneDescriptor to catch queryset.model.DoesNotExist so that hasattr/getattr on a missing reverse OneToOne returns False/None instead of raising.

Adds a regression test to HistoricOneToOneFieldTest covering the missing-relation case.

```
from onkol.apps.users.models import A
from django.utils.timezone import now

# Create object
a = A.objects.create()

# Get historical version at current time
historic_a = a.history.as_of(now())

# Assertions
assert not hasattr(a, 'b')
assert not hasattr(historic_a, 'b')
```

## Types of changes

## Checklist:

### DIFF
--- a/simple_history/models.py
+++ b/simple_history/models.py
@@ -980,6 +980,34 @@ class HistoricReverseOneToOneDescriptor(
     def get_related_model(self):
         return self.related.related_model
 
+    def __get__(self, instance, cls=None):
+        if instance is None:
+            return self
+
+        try:
+            rel_obj = self.related.get_cached_value(instance)
+        except KeyError:
+            if instance.pk is None:
+                rel_obj = None
+            else:
+                filter_args = self.related.field.get_forward_related_filter(instance)
+                queryset = self.get_queryset(instance=instance)
+                try:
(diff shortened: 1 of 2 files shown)


JordanHyatt:
Seems a little unfortunate we have to re-write the entire `__get__` method.

Especially because I recently submitted this PR for Django that would instantly break this patch:
https://github.com/django/django/pull/21415 (OneToOneField.related_default).

Additionally, it looks like Django has redesigned the `ReverseOneToOneDescriptor.__get__` method anyway to no longer catch `DoesNotExist` which may fix this problem. 

Not sure the best way forward here

PR #1594, opened 2026-03-23
Fix issue #341: Warn when using HistoricalRecords on an abstract c

JackDKobe:
Fixes good-first-issue #341


ddabble:
[Same goes for this](https://github.com/django-commons/django-simple-history/pull/1595#issuecomment-4754233649)

